PR #2, opened 2024-10-07
Update pbr is_integer

bigsmile74:
is_integer was defined twice.
pulled the first outside its parent function.
removed the second one. 
optimized the function


stangri:
Thanks for the improvement, I'll merge this but then reintroduce your more compact code as a function within the `output()` as I use the `output` function in various other scripts and would like to maintain portability, wherever `is_integer` is defined globally or not.

stangri:
oh, one more thing I kept forgetting about, please make future PRs (if any) against 1.1.7 branch, not main branch. ;)

PR #3, opened 2024-10-21
Update pbr: simplified filter_options

bigsmile74:
rewrote: filter_options
updated: is_ipv4 to support /netmask
updated: is_family_mismatch and nftset that used is_ipv4_netmask
dropped: is_ipv4_netmask

please test :)

### DIFF
--- a/files/etc/init.d/pbr
+++ b/files/etc/init.d/pbr
@@ -188,90 +188,13 @@ pbr_get_gateway6() {
 	fi
 	eval "$1"='$gw'
 }
-filter_options() {
-	local opt="$1" value="$2"
-	local i _ret=
-
-	case "$opt" in
-		phys_dev)
-			for i in $value; do
-				if is_phys_dev "$i"; then
-					_ret="${_ret:+$_ret }$i"
-				fi
-			done
-		;;
-		phys_dev_negative)
-			for i in $value; do
-				if is_negation "$i" && is_phys_dev "${i:1}"; then
-					_ret="${_ret:+$_ret }$i"
-				fi
-			done
-		;;
-		mac_address)
-			for i in $value; do
-				if is_mac_address "$i"; then
-					_ret="${_ret:+$_ret }$i"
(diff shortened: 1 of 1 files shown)


files/etc/init.d/pbr
```diff
@@ -188,90 +188,13 @@ pbr_get_gateway6() {
...
-		*)
-			echo ''
-			return 1
-		;;
-	esac
-
+filter_options(){
+	local opt="$1" values="$2" v _ret
+	for v in $values; do
+		case "$opt" in *_negative) is_negation "$v" || continue;;esac
```
stangri, line 194: This is a very elegant code now with the is_ipv4 improvement, I wonder if you could replace the case with the `str_contains` function which already exists.

stangri: the comment applies to line 194, not sure why GH is displaying 4 lines there.

bigsmile74: Yes, that function would do fine as a replacement for the case statement.

bigsmile74: Looked at it in more depth, its almost a good replacement, but not in this case:

str_contains matches _negative in the whole string
the case statement only if it ends with _negative

str_contains "$opt" _negative && { is_negation "$v" || continue;}
case "$opt" in *_negative) is_negation "$v" || continue;;esac

You could create another function but if it is only used in this function, what would be the point.
str_endswith() { [ $# -eq 2 ] && [ "${1%$2}" != "$1" ];}
str_contains() { [ $# -eq 2 ] && [ "${1//$2}" != "$1" ];}

stangri: Given that the opt is restricted to a limited number of strings which would never start with _negagive and could only end with _negative, I feel that str_contains is safe to use here.

stangri: squashing was allowed already.

stangri:
Thank you!

PR #4, opened 2024-10-22
Update pbr: better is_phys_dev, drop is_phys_dev_quick

bigsmile74:
is_phys_dev checks for symlink in /sys/class/net that contains all physical network devices. Its ~ 450x faster, and now as quick as is_phys_dev_quick is. So I replaced that with is_phys_dev.

```
# time ./is_phys_dev_quick
real	0m 0.31s
user	0m 0.23s
sys	0m 0.07s
```
```
# time ./is_phys_dev
real	0m 0.31s
user	0m 0.27s
sys	0m 0.04s
```


stangri:
That's amazing, thank you!

PR #5, opened 2024-10-23
Update pbr: is_domain

bigsmile74:
I pulled the regex from somewhere, but I did not save the link. It works with everything I threw against it. Maybe you have some edge cases for testing.

### DIFF
--- a/files/etc/init.d/pbr
+++ b/files/etc/init.d/pbr
@@ -300,7 +300,7 @@ uci_get_device() { uci_get 'network' "$1" 'device' || uci_get 'network' "$1" 'de
 uci_get_protocol() { uci_get 'network' "$1" 'proto'; }
 is_default_dev() { [ "$1" = "$(ip -4 r | grep -m1 'dev' | grep -Eso 'dev [^ ]*' | awk '{print $2}')" ]; }
 is_disabled_interface() { [ "$(uci_get 'network' "$1" 'disabled')" = '1' ]; }
-is_domain() { ! is_ipv6 "$1" && ! is_mac_address "$1" && ! is_phys_dev "$1" && str_contains "$1" '[a-zA-Z]'; }
+is_domain(){ echo "$1" | grep -qE '^([a-zA-Z0-9][a-zA-Z0-9-]{0,61}[a-zA-Z0-9]\.)+[a-zA-Z]{2,}$';}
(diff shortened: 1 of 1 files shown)


stangri:
Thanks for this. Could you please confirm if this regex works the same with busybox grep and standalone grep?

It seems to fail on the TLDs tho:
```sh
edge620 in /tmp # is_domain at && echo yes
edge620 in /tmp #
```

bigsmile74:
Correct, but that isn't a domain you can visit. So you want to be able to redirect <all_domains>.at ? This should do it:
```
is_domain(){ echo "$1" | grep -qE '^([a-zA-Z0-9][a-zA-Z0-9-]{0,61}[a-zA-Z0-9]\.)*[a-zA-Z]{2,}$';}
```
For future reference: https://www.rfc-editor.org/rfc/rfc1035
```
Note that while upper and lower case letters are allowed in domain
names, no significance is attached to the case.  That is, two names with
the same spelling but different case are to be treated as if identical.

The labels must follow the rules for ARPANET host names.  They must
start with a letter, end with a letter or digit, and have as interior
characters only letters, digits, and hyphen.  There are also some
restrictions on the length.  Labels must be 63 characters or less.
```

Considering grep, if its in busybox grep its in the full grep, otherwise scripts would fail.
```
# grep --help 2>&1 | grep -- '-[qE]';readlink $(which grep)
	-q	Quiet. Return 0 if PATTERN is found, 1 otherwise
	-E	PATTERN is an extended regexp
/bin/busybox
```
```
# grep --help 2>&1 | grep -- '-[qE]';readlink $(which grep)
  -E, --extended-regexp     PATTERNS are extended regular expressions
  -q, --quiet, --silent     suppress all normal output
if any error occurs and -q is not given, the exit status is 2.
/usr/libexec/grep-gnu
```

stangri:
> Correct, but that isn't a domain you can visit.

It's a valid entry for dnsmasq nft set entries tho and that's what I'm actually trying to filter here. 

Thank you for this!

stangri:
@bigsmile74 there's an issue:

```
edge620 in ~ # is_domain(){ echo "$1" | grep -qE '^([a-zA-Z0-9][a-zA-Z0-9-]{0,61}[a-zA-Z0-9]\.)*[a-zA-Z]{2,}$'; }
edge620 in ~ # is_domain x.com && echo yes
edge620 in ~ # is_domain t.co && echo yes
```

PR #6, opened 2024-10-23
Update pbr: is_ovpn

bigsmile74:
I read up on [Universal TUN/TAP device driver](https://www.kernel.org/doc/html/v6.12-rc4/networking/tuntap.html) I could not make out if tun_flags is only for tun and tap maybe has tap_flags?

In case that it does:
`is_ovpn(){ local f=$(uci get network.$1.device 2>/dev/null);[ -f "/sys/class/net/$f/${f%%[0-9]*}_flags" ];}`

### DIFF
--- a/files/etc/init.d/pbr
+++ b/files/etc/init.d/pbr
@@ -322,7 +322,7 @@ is_negation() { [ "${1:0:1}" = '!' ]; }
 is_netifd_table() { grep -q "ip.table.*$1" /etc/config/network; }
 is_netifd_table_interface() { local iface="$1"; [ "$(uci_get 'network' "$iface" 'ip4table')" = "${packageName}_${iface%6}" ]; }
 is_oc() { local p; network_get_protocol p "$1"; [ "${p:0:11}" = "openconnect" ]; }
-is_ovpn() { local d; uci_get_device d "$1"; [ "${d:0:3}" = "tun" ] || [ "${d:0:3}" = "tap" ] || [ -f "/sys/devices/virtual/net/${d}/tun_flags" ]; }
+is_ovpn(){ [ -f "/sys/class/net/$(uci get network.$1.device 2>/dev/null)/tun_flags" ];}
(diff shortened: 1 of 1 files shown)


stangri:
Last I checked the tun_flags only existed if the tunnel was up, that's why there's a check for tap/tun in the device name, so that even if the OpenVPN tunnel is down, it can still be detected/enumerated.

If you can look into:
* is the OpenVPN package in 23.05 repo on the same version as in snapshots?
* do the tun_flags exist for both tap and tun devices when the tunnel is down?

We can then merge this.

I have removed the extra (second) definition of `uci_get_device()` tho.

bigsmile74:
These are from SNAPSHOT:
openvpn-mbedtls - 2.6.12-r1 - Open source VPN solution using mbedTLS
openvpn-openssl - 2.6.12-r1 - Open source VPN solution using OpenSSL
openvpn-wolfssl - 2.6.12-r1 - Open source VPN solution using WolfSSL

These are from OpenWRT 23.05.5:
openvpn-mbedtls - 2.5.8-5 - Open source VPN solution using mbedTLS
openvpn-openssl - 2.5.8-5 - Open source VPN solution using OpenSSL
openvpn-wolfssl - 2.5.8-5 - Open source VPN solution using WolfSSL

I'm currently looking into tun/tap, but its not only tun0/tun_flags that disappears, its the whole folder tun0.
We might need this later:
```
up: rx and tx counters increase
tun0/statistics/rx_bytes:3232	> increase
tun0/statistics/rx_packets:54	> increase
tun0/statistics/tx_bytes:6788	> increase
tun0/statistics/tx_packets:112	> increase

grace period: rx_bytes/packets stall
tun0/statistics/rx_bytes:3232	< stall
tun0/statistics/rx_packets:54	< stall
tun0/statistics/tx_bytes:6788	> increase
tun0/statistics/tx_packets:112	> increase

down: grace period tx_dropped: increases periodically
tun0/statistics/tx_dropped:6238	< increase (every ~ 0.5 seconds)
tun0/statistics/rx_bytes:3232	< stall
tun0/statistics/rx_packets:54	< stall
tun0/statistics/tx_bytes:6788	> stall
tun0/statistics/tx_packets:112	> stall
```

stangri:
If the check for device name (tap/tun) has to stay, do you still want to change the `is_ovpn()` function in any way then?

bigsmile74:
Yes this one, will take some more time to do properly. You can close it or leave it open, I'll pick it up when I have more time.

PR #7, opened 2024-10-23
Update pbr: warn_dhcp_force

bigsmile74:
Display a warning if user has dhcp.force = 0 (disabled). If disabled a request is put out to identify other DHCP-servers on the network. This takes up time (about 3s ¬ 6s depending on start/stop/restart) and is uncommon for most users.

I could not find a good place to put it because the code is still unfamiliar., so its added to restart to give an idea how it would function.

### DIFF
--- a/files/etc/init.d/pbr
+++ b/files/etc/init.d/pbr
@@ -25,6 +25,8 @@ readonly _OKB_='\033[1;34m\xe2\x9c\x93\033[0m'
 readonly __OKB__='\033[1;34m[\xe2\x9c\x93]\033[0m'
 readonly _FAIL_='\033[0;31m\xe2\x9c\x97\033[0m'
 readonly __FAIL__='\033[0;31m[\xe2\x9c\x97]\033[0m'
+readonly _WARN_='\033[0;33m!\033[0m'
+readonly __WARN__='\033[0;33m[!]\033[0m'
 readonly _ERROR_='\033[0;31mERROR\033[0m'
 readonly _WARNING_='\033[0;33mWARNING\033[0m'
 readonly ip_full='/usr/libexec/ip-full'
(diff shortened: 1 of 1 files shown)


stangri:
Thanks again for your amazing contributions!

On this one I have some comments:
* I think it's a better idea to run this check as a part of the `_system_health_check()` sub-function of `load_environment()` function.
* I don't think the code needs to be wrapped in its own separate function, can be just part of `_system_health_check()`.
* There's already a `readonly _WARNING_` variable, why not use it? Please see other warnings in the code for reference.
* Please add a line to the `get_text()` with a warning ID similar (in naming) to the existing ones.
* Please use `state add 'errorSummary'` to add warning to the run-time state by the warning ID you've created above.
* Please make sure to bump `packageCompat` by one after completing two steps above.

I can take care of adding a new ID and compat-bumping the luci app, unless you want to take a shot at it too.

I'd also like to bump the PKG_RELEASE after all your contributions are merged, if you want to create a PR for OpenWrt packages repo (from this repo) with these changes (let me know if I need to set up permissions for you on this repo for this) for like Hacktoberfest or other reasons, let me know, otherwise I'll do the PR and merge changes into OpenWrt repo myself.

bigsmile74:
> I can take care of adding a new ID and compat-bumping the luci app, unless you want to take a shot at it too.
> 
> I'd also like to bump the PKG_RELEASE after all your contributions are merged, if you want to create a PR for OpenWrt packages repo (from this repo) with these changes (let me know if I need to set up permissions for you on this repo for this) for like Hacktoberfest or other reasons, let me know, otherwise I'll do the PR and merge changes into OpenWrt repo myself.

Thank you for the trust, but I'll keep digging through the code and making PR's for now as my time is limited.

stangri:
If you don't mind I can close this and migrate your code into `_system_health_check()` and add the necessary text to both principal and luci packages then.

PR #8, opened 2024-10-25
Update pbr: process_url

bigsmile74:
UNTESTED:

Split process_url into two parts:

- set_dl that figures out the best download tool
- process_url for downloading

I did not know what to put in my configuration of PBR to test it on short notice. So this is untested. Maybe you have suggestions for a couple of triggers.

Also I traded in the sed regex for xargs which I interpreted as compatible, but needs to be tested. Download is redirected to stdout where its split into arguments by xargs. (no temp file needed anymore)

Please if you think of a good error text, I could not.

### DIFF
--- a/files/etc/init.d/pbr
+++ b/files/etc/init.d/pbr
@@ -421,30 +421,30 @@ get_text() {
 	esac
 	echo "$r"
 }
-set_dl(){
+set_dl() {
 	[ -x "$wget" ] && {
 		dl_schemes='ftp http'
-		dl_command='$wget -q --no-check-certificate'
+		dl_command="$wget -q --no-check-certificate"
 		dl_flag='-O'
 	}
 	[ -x "$fetch" ] && {
 		[ -f "$fetchssl" ] && dl_schemes='ftp http https' || dl_schemes='ftp http'
-		dl_command='$fetch -q --no-check-certificate'
+		dl_command="$fetch -q --no-check-certificate"
 		dl_flag='-O'
 	}
 	[ -x "$wgetssl" ] && {
 		dl_schemes='ftp http https'
-		dl_command='$wgetssl -q --no-check-certificate'
+		dl_command="$wgetssl -q --no-check-certificate"
 		dl_flag='-O'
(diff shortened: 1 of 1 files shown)


files/etc/init.d/pbr
```diff
@@ -417,45 +421,37 @@ get_text() {
...
-	elif $dl_command "$url" "$dl_flag" "$dl_temp_file" 2>/dev/null; then
-		sed 'N;s/\n/ /;s/\s\+/ /g;' "$dl_temp_file"
-	else
-		state add 'errorSummary' 'errorDownloadUrl' "$url"
-	fi
-	rm -f "$dl_temp_file"
+set_dl(){
+	[ -x "$wget" ] && {
+		dl_schemes='ftp http'
+		dl_command='$wget -q --no-check-certificate'
```
stangri, line 427: Please use double quotes for dl_command expressions.

bigsmile74: Yes, otherwise it does not work :rofl:

stangri:
I like the introduction of dl_schemes per each download utility and just checking it against the schema used in the download. 

I think it would make sense to incorporate the code from dl_set() function back into process_download tho, and declare `dl_schemes` as a local variable in process_download. 

If you have curl installed you can test with a local file containing some domains. 

I can bump compat variables and improve/bring the error string into luci app after merging PR.

bigsmile74:
(Would have gotten back sooner, but I'm swamped a the moment) Personally I would move dl_set into _system_health_check and check everything beforehand. Aborting then would prevent the system to become semi-configured.

stangri:
> (Would have gotten back sooner, but I'm swamped a the moment) Personally I would move dl_set into _system_health_check and check everything beforehand. Aborting then would prevent the system to become semi-configured.

Now that there is a `_system_health_check`, that might not be a bad idea, even at the expense of having to parse the pbr config to look for links on start. Let me sleep on this.

stangri:
closing as 1.1.7 branch is outdated.

PR #9, opened 2024-10-25
Update pbr: ouput

bigsmile74:
Optimized the output function as best I could. But I think there might be a better solution for this. I can only see the code, could you explain in more detail what it is exactly that you want to accomplish with this function? And what would you like to go where regarding console, logger, log file ,error file.

### DIFF
--- a/files/etc/init.d/pbr
+++ b/files/etc/init.d/pbr
@@ -141,27 +141,18 @@ quiet_mode() {
 	esac
 }
 output() {
-# Target verbosity level with the first parameter being an integer
-	is_integer() { case "$1" in ''|*[!0-9]*) return 1;; esac; }
-	local msg memmsg logmsg text
-	local sharedMemoryOutput="/dev/shm/$packageName-output"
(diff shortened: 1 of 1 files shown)


stangri:
Thank you for another contribution.

I definitely like parts of it, like getting rid of unfortunately named  `$memmsg` and `$logmsg` and using `$queue` instead of a long variable name, however:

* It's a personal preference, but I'm not a big fan of combining statements in curly braces after `&&`
* You've declared `local msg` twice
* The shellcheck should also complain about `case $1`
* I feel that the old logic would be broken in a case of `output '20'` or  `output '21'`, depending on your settings
* Would be nice to get rid of use of `wc -l` while we're at it

So while I'd like to steal some ideas and I don't want to discourage you from further contributions, I can't accept this new piece of code as is. Let me know if you're comfortable with me incorporating part of the code into `pbr` and my other packages or if you want to shellcheck it and keep at it.

stangri:
Superceeded by d13b71d6f61bc356048daf3fbe72249a6d6e64d3, thanks @bigsmile74

PR #15, opened 2024-11-01
Update pbr: ip-full check

bigsmile74:
You already had ip_full specified in the readonly section, so used that as base. Then implemented the comment error message.

Maybe its better to suggest a package to install as they contain the binary. Sometimes they are not the same.


stangri:
Thank you! I'll work on the error message.

PR #16, opened 2024-11-02
Update pbr: is_supported_protocol

bigsmile74:
source: https://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml
file: /etc/protocols
1. Grep with first argument at beginning of the line.
2. Sets default argument value to -, otherwise all lines would match.
3. Used -i to make it case insensitive, so ipv6 as well as IPv6 would match.
4. Used -q to make it quiet.


stangri:
Thank you!

PR #17, opened 2024-11-02
Update pbr: ipv4_leases_to_nftset

bigsmile74:
1. Used only awk
2. Passed $1 along for search
3. Started with an empty field separator
4. Then used printf to write the output


stangri:
Thank you!

PR #18, opened 2024-11-02
pbr: is_supported_interface update

egc112:
Present code logic will process an interface if it is listed as supported even if it is disabled.

To me it makes more sense that if an interface is disabled it will not be processed at all.

This patch should make that so.

If you agree then please review and implement, otherwise just discard and close


stangri:
Makes sense Erik, thank you, I hope the visit is going great!

PR #20, opened 2024-11-02
Update pbr: inline_set

bigsmile74:
1. Add spaces to begin and end while adding a comma
2. Remove leading @ or ! from every field
3. Remove leading space
4. Remove trailing space and echo back.

### DIFF
--- a/files/etc/init.d/pbr
+++ b/files/etc/init.d/pbr
@@ -187,15 +187,7 @@ filter_options() {
 	echo "$_ret"
 	return 0
 }
-inline_set() {
-	local value="$1" inline_set i
-	for i in $value; do
-		[ "${i:0:1}" = "!" ] && i=${i:1}
-		[ "${i:0:1}" = "@" ] && i=${i:1}
-		inline_set="${inline_set:+$inline_set, }$i"
-	done
-	echo "$inline_set"
-}
+inline_set(){ set " ${1// /, } ";set "${1// [@!]/ }";set "${1# }";echo "${1% }";}
 # shellcheck disable=SC2016
 is_bad_user_file_nft_call() { grep -q '"\$nft" list' "$1" || grep '"\$nft" -f' "$1"; }
 is_config_enabled() {
--- a/files/etc/init.d/pbr
+++ b/files/etc/init.d/pbr
@@ -187,7 +187,13 @@ filter_options() {
 	echo "$_ret"
 	return 0
 }
-inline_set(){ set " ${1// /, } ";set "${1// [@!]/ }";set "${1# }";echo "${1% }";}
+inline_set() {
+	local value="$1" inline_set i
+	for i in $value; do
+		inline_set="${inline_set:+$inline_set, }${i#[@!]}"
+	done
+	echo "$inline_set"
+}
 # shellcheck disable=SC2016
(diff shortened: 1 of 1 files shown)


files/etc/init.d/pbr
```diff
@@ -187,15 +187,7 @@ filter_options() {
...
-inline_set() {
-	local value="$1" inline_set i
-	for i in $value; do
-		[ "${i:0:1}" = "!" ] && i=${i:1}
-		[ "${i:0:1}" = "@" ] && i=${i:1}
-		inline_set="${inline_set:+$inline_set, }$i"
-	done
-	echo "$inline_set"
-}
+inline_set(){ set " ${1// /, } ";set "${1// [@!]/ }";set "${1# }";echo "${1% }";}
```
stangri, line 190: Could you please elaborate on the need to insert the trailing space? Like why not `inline_set() { set " ${1// /, }"; set "${1// [@\!]/ }"; echo "${1# }"; }`.

Also, I have a concern that on bad input, any double (or triple) spaces will turn into ` ,  ,  ,` which might mess up the set. I don't know of any elegant parameter expansion/substitution to take care of that, do you?

bigsmile74:
Afaik variable substitution does not handle our variable white space. (good catch,trailing space is obsolete and can be skipped) 

### So that leaves:
1. shell: field iteration with variable substitution for fields leading with @ !
```
inline_set() {
	local value="$1" inline_set i
	for i in $value; do
		inline_set="${inline_set:+$inline_set, }${i#[@!]}"
	done
	echo "$inline_set"
}
```
2. awk: no field iteration, replace white space(1 step), remove leading @ ! and set field separator
```
inline_set(){ echo " $1 "|awk 'BEGIN{OFS=", "}{gsub(/\s[!@]/,"");$1=$1;print}';}
```
3. sed: no field iteration, replace white space(4 steps), remove leading @ ! and set field separator
```
inline_set(){ echo " $1 "|sed 's/\s[@!]/ /g;s/\s\+/ /g;s/^ //;s/ $//;s/ /, /g';}
```
\*\* **_not applicable to your code but since i was testing:_** \*\*
4. same awk code out of loop eg: { for i in 0 to 10000; do something;done} total output piped through awk
5. same sed code out of loop eg: { for i in 0 to 10000; do something;done } total output piped through sed

### \# 1. Clearly wins
I timed a loop of 10k iterations and although awk out of loop wins, that only counts for larger data sets.
So 1. clearly wins as this is a more realistic approach and is 150% faster than 0. the former code.

#### Results 10k loop:
```
+ time ./0 # original shell code
real	0m 0.43s
user	0m 0.42s
sys	0m 0.00s
+ time ./1 # simplified shell code
real	0m 0.17s
user	0m 0.17s
sys	0m 0.00s
+ time ./2 # awk in loop
real	0m 14.27s
user	0m 6.77s
sys	0m 3.02s
+ time ./3 # sed in loop
real	0m 15.89s
user	0m 8.35s
sys	0m 3.41s
+ time ./4 # awk out of loop
real	0m 0.11s
user	0m 0.13s
sys	0m 0.00s
+ time ./5 # sed out of loop
real	0m 0.15s
user	0m 0.18s
sys	0m 0.00s
```

stangri:
Thank you for stress-testing this. Would have been nice to have it as one-liner, but if first option wins on both speed and safety, I agree it's the best.

PR #23, opened 2024-11-04
Update pbr: remove sync slowdown ?

bigsmile74:
What's the reason to use sync? Most of it runs in-memory or on a tmpfs. If you let Linux handle it will save about ~5 seconds in a restart.

I only see it used in the following init files with a good reason:
* boot: to cement initial config generation.
* done: after removal of sysupgrade (substantial file size)
* uhttpd to cement the certificate
* umount to cement pending writes

I experience no issues by removing them.

before:
```
# time /etc/init.d/pbr restart
real	0m 7.62s
user	0m 1.31s
sys	0m 0.37s
```
after:
```
# time ./pbr restart
real	0m 2.37s
user	0m 1.33s
sys	0m 0.40s
```

### DIFF
--- a/files/etc/init.d/pbr
+++ b/files/etc/init.d/pbr
@@ -814,7 +814,6 @@ cleanup_rt_tables() {
 	for i in $(grep -oh "${ipTablePrefix}_.*" "$rtTablesFile"); do
 		! is_netifd_table "$i" && sed -i "/${i}/d" "$rtTablesFile"
 	done
-	sync
 }
 
 cleanup_main_chains() {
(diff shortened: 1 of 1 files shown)


bigsmile74:
Also and I did not mention this, there is the penalty of wear-leveling if you run from `NAND/NOR`  flash. So by letting those writes accumulate in cache it would only write the end result. If you still want a sync, I would do it before the script exits and in the background eg: `sync &` then exit.

stangri:
I believe when I first implemented the rt_tables updates there were issues with people ending up with corrupted tables, that's why the sync was implemented. The issue is if I accept the PR and start releasing new binaries, it may be a while until we get a corruption report. Again, let me sleep on this.

stangri:
closing as 1.1.7 branch is otudated

PR #24, opened 2024-11-05
Update pbr: made "enabled" compatible with initrc and service (depends on second patch in luci-app-pbr)

bigsmile74:
This makes PBR compatible with initrc and service. but depends on another [patch](https://github.com/stangri/luci-app-pbr/pull/1/commits/0e5a59f3489a6f7e26e9e21d1eafea67357c8b8b) for luci-app-pbr to make it function.
Changes:
* new readonly variable startOnBoot
* removed variable enabled (obsolete)
* changed <del>removed</del> function is_enabled
* removed config_get_bool enable (obsolete)
* changed function load_environment
* <del>removed $enabled entries 6x</del>
* <del>new function: enable() to enable the service</del>
* <del>new function: disable() to disable the service</del>

### DIFF
--- a/files/etc/init.d/pbr
+++ b/files/etc/init.d/pbr
@@ -2038,7 +2038,8 @@ start_service() {
 	fi
 
 	procd_open_instance 'main'
-	procd_set_param command /bin/true
+	procd_set_param command /bin/sleep '10000d' #temporary workaround
+	procd_set_param pidfile $packageLockFile
 	procd_set_param stdout 1
 	procd_set_param stderr 1
 	procd_open_data
@@ -2133,7 +2134,6 @@ service_started() {
 	fi
 	state print 'errorSummary'
 	state print 'warningSummary'
-	touch "$packageLockFile"
 	if [ -n "$errorSummary" ]; then
 		return 2
 	elif [ -n "$warningSummary" ]; then
@@ -2200,7 +2200,6 @@ stop_service() {
 			output "$serviceName (nft mode) stopped "; output_okn;
 		fi
 	fi
-	rm -f "$packageLockFile"
 }
 
 version() { echo "$PKG_VERSION"; }
@@ -2401,5 +2400,3 @@ load_validate_include() {
 		'path:file' \
 		'enabled:bool:0'
 }
-
-service_running(){ [ -f "$packageLockFile" ];}
--- a/files/etc/init.d/pbr
+++ b/files/etc/init.d/pbr
@@ -48,7 +48,6 @@ readonly rtTablesFile='/etc/iproute2/rt_tables'
 
 # package config options
 procd_boot_timeout=
-enabled=
 fw_mask=
 icmp_interface=
 ignored_interface=
@@ -298,7 +297,6 @@ uci_get_listen_port() {
 }
 
 # luci app specific
-is_enabled() { uci_get "$1" 'config' 'enabled'; }
 is_running_nft_file() { [ -s "$nftPermFile" ]; }
 is_running_nft() { "$nft" list table inet fw4 | grep chain | grep -q pbr_mark_ >/dev/null 2>&1; }
 check_nft() { [ -x "$nft" ]; }
@@ -347,7 +345,6 @@ get_text() {
 		errorConfigValidation) r="Config ($packageConfigFile) validation failure!";;
 		errorNoNft) r="Resolver set support (${resolver_set}) requires nftables, but nft binary cannot be found!";;
 		errorResolverNotSupported) r="Resolver set (${resolver_set}) is not supported on this system!";;
-		errorServiceDisabled) r="The ${packageName} service is currently disabled!";;
 		errorNoWanGateway) r="The ${serviceName} service failed to discover WAN gateway!";;
 		errorNoWanInterface) r="The %s interface not found, you need to set the 'pbr.config.procd_wan_interface' option!";;
 		errorNoWanInterfaceHint) r="Refer to https://docs.openwrt.melmac.net/pbr/#procd_wan_interface.";;
@@ -442,7 +439,6 @@ load_package_config() {
 	local param="$1"
 	local user_file_check_result i
 	config_load "$packageName"
-	config_get_bool enabled                   'config' 'enabled' '0'
 	config_get      fw_mask                   'config' 'fw_mask' 'ff0000'
 	config_get      icmp_interface            'config' 'icmp_interface'
 	config_get      ignored_interface         'config' 'ignored_interface'
@@ -552,11 +548,6 @@ load_environment() {
 		on_boot|on_start)
 			output 1 "Loading environment ($param) "
 			load_package_config "$param"
-			if [ "$enabled" -eq '0' ]; then
-				output 1 "$_FAIL_\n"
-				state add 'errorSummary' 'errorServiceDisabled'
-				return 1
-			fi
 			if [ -n "$validation_result" ] && [ "$validation_result" != '0' ]; then
 				output 1 "$_FAIL_\n"
 				output "${_ERROR_}: The $packageName config validation failed!\n"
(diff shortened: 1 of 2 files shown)


bigsmile74:
Link to [luci-pbr-patch](https://github.com/stangri/luci-app-pbr/pull/1/commits/0e5a59f3489a6f7e26e9e21d1eafea67357c8b8b)

stangri:
I prefer to keep the enabled control in the uci config file, as otherwise the service is enabled on the install and starts/runs immediately after.

bigsmile74:
If that is the only issue, I will add disabling the service after installation. _(As it is now, it's not compatible with the OpenWrt service standard which people expect it to be.)_

Could you look into the "removed $enabled entries 6x", I'm unsure if that was accurate.

files/etc/init.d/pbr
```diff
@@ -1466,8 +1457,6 @@ dns_policy_process() {
 policy_process() {
 	local i j uid="$1"
 
-	[ "$enabled" -gt '0' ] || return 0
```
stangri, line 1469: this is the check for the policy enabled flag, not for service flag.

files/etc/init.d/pbr
```diff
@@ -1930,7 +1919,6 @@ process_interface() {
 
 user_file_process() {
 	local shellBin="${SHELL:-/bin/ash}"
-	[ "$enabled" -gt '0' ] || return 0
```
stangri, line 1933: this is the check for the custom user file enabled flag, not for service flag.

files/etc/init.d/pbr
```diff
@@ -2202,12 +2190,10 @@ stop_service() {
 	resolver 'store_hash'
 	resolver 'cleanup_all'
 	resolver 'compare_hash' && resolver 'restart'
-	if [ "$enabled" -ne '0' ]; then
```
stangri, line 2205: this just suppresses the output when the service is stopped after it's been disabled (ie: uninstall scenario).

stangri:
OK, so I'm not convinced this is the way to go, because setting the pbr to run at boot-time is not the same as enabling pbr to run period. For example, someone may want to disable pbr altogether (for testing or other purposes), but disabling pbr to run at boot (which is what this PR does) will not prevent pbr from starting on say, firewall restart/reload or an interface update. So in short, I don't think you can equate boot start-up to enabling/disabling service.

Anyways, let me sleep on it, but a few things to think over if this were to be accepted:
1. The documentation would also need to be updated, as it has instructions on enabling the package.
2. I'd prefer to keep the is_enabled() function, just modify it to reflect the new enabled status check.
3. The luci rpcd script sources the init file, so all the shared shell functions should only be in the init file, not duplicated in the rpcd script. The rpcd script functions should only be applicable to the rpcd script itself.

bigsmile74:
Oh, I missed those "local enabled", something felt off about the number of patches.

I redid the patch, its now not that different from what it was, except for the compatibility with the service command.
The enabled option has become obsolete, while maintaining the same functionality.

stangri:
> this should also fix the installation error because the service does not get enabled

Sorry, could you please elaborate on that?

Also, I appreciate you fixing the issues I pointed out, however it didn't address the biggest problem -- equating enabling service to run at boot-time with enabling the service in more general terms.

bigsmile74:
I was in the middle of editing my previous post when I saw your comment under `EDIT2`.

stangri:
> Actually that is exactly how it works, see this snippet from: /etc/rc.common

It's definitely not how it works for multiple packages, starting (as in the oldest) with miniupnpd/adblock and most of my packages. The unfortunately named enabled/disabled functions in rc.common were most likely created for the System->Startup luci page/functions, but definitely not all packages from OpenWrt packages repo rely just on that softlink to determine if they are enabled or not.

Considering different use-case scenarios, if the enabled option in uci config is replaced with the boot-up link:
1. How would you handle the case where someone doesn't want pbr started on boot, but wants to start pbr manually later? 
2. How would you handle the upgrade to the newer version -- how would you retain the previously enabled/disabled status? 
3. In the case the package needs to be uninstalled to be installed later (so let's say there's a reboot between previous uninstall and the next install), how would you retain the previously enabled/disabled status?

bigsmile74:
> but definitely not all packages from OpenWrt packages repo rely just on that softlink to determine if they are enabled or not.

Correct, but they supply their own `enabled()` `enable()` `disable()` functions. Unless these are defined by the developer, the default behavior if `rc.common` is sourced, is that the `symlink` determines if a service is enabled/disabled. The service command just requests the enabled status from each initscript as defined in `/sbin/service`
```
       for service in /etc/init.d/* ; do
                boot="$($service enabled && echo "enabled" || echo "disabled" )"
```
.

> Considering different use-case scenarios, if the enabled option in uci config is replaced with the boot-up link:
> 1. How would you handle the case where someone doesn't want pbr started on boot, but wants to start pbr manually later?

By removing the extra constraint that the service must be enabled. Afaict that is already defined in this file:
`/usr/share/pbr/firewall.include`
```
#!/bin/sh
if [ -x /etc/init.d/pbr ] && /etc/init.d/pbr enabled; then
	logger -t "pbr" "Sending reload signal to pbr due to firewall action: $ACTION"
	/etc/init.d/pbr on_firewall_reload "$ACTION"
fi
```
 
> 2. How would you handle the upgrade to the newer version -- how would you retain the previously enabled/disabled status?

I think informing the user would be best because the code becomes obsolete after, but if this needs to be automated:
```
local enabled=$(uci get pbr.config.enabled 2>/dev/null)
[ -n "$enabled" ] && {
  [ $enabled -eq 1 ] && service pbr enable || service pbr disable
  [ $? -eq 0 ] && uci delete pbr.config.enabled || echo "error: unable to delete pbr.config.enabled"
}
```
> 3. In the case the package needs to be uninstalled to be installed later (so let's say there's a reboot between previous uninstall and the next install), how would you retain the previously enabled/disabled status

`It cannot`, but that also means the device is in maintenance mode and the user should be aware of that, unless I am missing something? Do you think this is a common occurrence?, because it can always be added.

bigsmile74:
That's a very good point, I could export a variable or write a tmpfile, that could get picked up on re-install. Let me do some more coding and testing.

stangri:
> That's a very good point, I could export a variable or write a tmpfile, that could get picked up on re-install. Let me do some more coding and testing.

Or we could keep the enabled option in the uci config. ;)

stangri:
So in the case uninstall/reboot/reinstall, the enabled status is gone again.

bigsmile74:
You caught me in the middle of an update :) Does your comment still stand?

stangri:
OK, maybe we need to roll back a little, what user experience problem are you trying to solve here?

bigsmile74:
Making `PBR` compatible with the `service` command _(some users run without luci)_

Using the `service` command:
```
service | grep pbr
/etc/init.d/pbr               	  disabled	   stopped
```
Starting service
```
# service pbr start >/dev/null 2>&1
# service | grep pbr
/etc/init.d/pbr               	  disabled	   running
```
Enabling service
```
# service pbr enable
# service | grep pbr
/etc/init.d/pbr               	   enabled	   running
```
Stopping service
```
# service pbr stop >/dev/null 2>&1
# service | grep pbr
/etc/init.d/pbr               	   enabled	   stopped
```
Disabling service
```
# service pbr disable
# service | grep pbr
/etc/init.d/pbr               	  disabled	   stopped
```

stangri:
It already is compatible with the service command and works as intended. The service enable/disable commands affect the boot-up time start only and that's how it currently works.

Also, the uci config option can be set from the CLI (without luci) and the package even informs the user with the exact commands to run to enable package.

It would be trivial to overwrite the rc.common's enable/disable/enabled functions to perform as they do and also affect/check the uci config option, however I'm not sure if that is desirable. 

There are other packages where the boot-up start and generally enabling/disabling package to run are separated, so it's not like pbr is setting the precedent.

bigsmile74:
Just submitted [PR16992](https://github.com/openwrt/openwrt/pull/16992/files) for OpenWrt that lets '/sbin/service' support services that:
* do not have an active process under procd _(like pbr)_
* do not use procd

Lets park this PR for now.

stangri:
Good luck with the PR in openwrt/openwrt.

bigsmile74:
Thank you, lets hope they adopt it.

PR #25, opened 2024-11-09
pbr: ip rule delete and sport patch

egc112:
This patch should solve a problem where the ip rules, in particularly the ipv6 rules are not properly deleted One problem is that `interface_routing 'destroy'` does not send the priority number so that the rules are not deleted at all.

For non netifd tables the IPv4 rules are flushed later on but not for IPv6.

By fixing this we should be able to use the same priority number for ipv6 instead of using `$((priority-1))`

Second problem was that wg servers sport where not set and destroyed properly for ipv6 this should also be fixed.

Tested on my R7800 build 23.05.5 nss build

Please have a look and consider implementing

Thanks for all your great work

### DIFF
--- a/files/etc/init.d/pbr
+++ b/files/etc/init.d/pbr
@@ -1560,7 +1560,7 @@ interface_routing() {
 				if [ -n "$ipv6_enabled" ]; then
 					ipv6_error=0
 					ip -6 rule del table "$tid" prio "$priority" >/dev/null 2>&1
-					try ip -6 rule add fwmark "${mark}/${fw_mask}" table "$tid" priority "$((priority-1))" || ipv6_error=1
+					try ip -6 rule add fwmark "${mark}/${fw_mask}" table "$tid" priority "$priority" || ipv6_error=1
 				fi
 			else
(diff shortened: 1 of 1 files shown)


files/etc/init.d/pbr
```diff
@@ -1773,15 +1783,15 @@ process_interface() {
 						ip rule add sport "$listen_port" table "pbr_${wanIface4}" >/dev/null 2>&1
 					fi
 					if [ -n "$ipv6_enabled" ] && [ -n "$wanIface6" ]; then
-						ip rule del sport "$listen_port" table "pbr_${wanIface6}" >/dev/null 2>&1
-						ip rule add sport "$listen_port" table "pbr_${wanIface6}" >/dev/null 2>&1
+						ip -6 rule del sport "$listen_port" table "pbr_${wanIface4}" >/dev/null 2>&1
```
stangri, line 1786: Does that work for wan/wan6? Maybe deletion should be duplicated for both `wanIface4` and `wanIface6`

stangri:
Merged with thanks!

PR #29, opened 2025-02-04
pbr: Ignore WireGuard Server if on ignored list

egc112:
Presently a WireGuard server is automatically ignored and a rule is automatically created to route the port out of the WAN, very convenient for a lot of users.

However if a user does not want this there is no way to disable this. See: https://forum.openwrt.org/t/policy-based-routing-pbr-package-discussion/140639/1965?u=egc and following posts.

With this PR the rule for routing the WG server port out of the WAN will not be made if the WG server is on the ignored list.

A simple solution but this changes previous behaviour, so at least the help text should be updated to reflect this.

I will also make a PR for changing the help text in luci-app-pbr.  
Probably the read.me should also be updated.

As this PR changes previous behaviour, I am not sure if this is the optimal solution, so feel free to change/improve  or discard this pull request 😊

Tested on R7800 running 24.10

Signed-off-by: Erik Conijn <[egc112@msn.com](mailto:egc112@msn.com)>


stangri:
Thank you Erik!

PR #43, opened 2025-09-14
pbr: fix install file pbr.user.dnsprefetch

pesa1234:
bump to r28


stangri:
Thank you for spotting this, merged with gratitude!

PR #44, opened 2025-09-18
Update pbr.dnsprefetch

betonmischer86:
Update pbr.dnsprefetch due to 38cc0b6.


stangri:
Oh, sorry, completely forgot you were using that variable!

PR #49, opened 2025-10-11
reorder config options, update config file

betonmischer86:
Per our [discussion](https://github.com/stangri/docs.openwrt.melmac.ca/issues/15#issuecomment-3353936191), update `/etc/config/pbr` to include all options that have a hardcoded default. Also reorder entries in load_package_config() alphabetically, except the `netifd_` options, which are WIP. I hope nothing got misplaced.


stangri:
Hey @betonmischer86 sorry about the delay in responding. Now that you (and copilot) brought it up, I can't recall the reason for `procd_reload_delay` change to '0'. I'll merge this as is, but may revert the defaults at some point in a future to prevent simultaneous reloads when interfaces start to come up.

PR #53, opened 2025-11-09
fix: preserve established connections on service restart

brauliobo:
Add connection tracking rules to preserve established/related connections when the PBR service restarts. This prevents existing SSH sessions from being broken when routing rules are flushed and recreated.

The fix adds rules at the top of prerouting and output chains that check for established/related connection state and return early, allowing those packets to continue using their existing routing table.

### DIFF
--- a/files/etc/init.d/pbr
+++ b/files/etc/init.d/pbr
@@ -815,6 +815,9 @@ nft_file() {
 				mkdir -p "${i%/*}"
 			done
 			{ echo '#!/usr/sbin/nft -f'; echo ''; } > "$nftTempFile"
+			# Preserve established/related connections that already have a routing mark to prevent breaking existing sessions (e.g., SSH)
+			echo "add rule inet $nftTable ${nftPrefix}_prerouting ct state established,related meta mark & $fw_mask != 0 return" >> "$nftTempFile"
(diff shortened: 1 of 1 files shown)


stangri:
Thanks for your contribution @brauliobo. As the `pbr` isn't the only package for OpenWrt marking packets, I wonder if this has been tested for potential unexpected behaviour from fw4 when SQM or other packet-marking packages are installed?

egc112:
For my understanding is the problem you are encountering that existing SSH connections are broken and you have to login again after PBR (re)starts?

The main chains already have a rule on top preserving related and established traffic, adding such a rule to the prerouting chain might have unwanted side effects but I am not an expert on nftables as I grew up with iptables and not with nftables

brauliobo:
Hi @stangri, yes I'm using SQM and nft sets with domain-based rules. I'm new to OpenWRT development and not able to test in detail. I have 3 WAN and 4 WG interfaces, so I'm testing rules to ping with each of them. I'm also using MWAN3, but I've set PBR a higher priority than it (lower `uplink_ip_rules_priority` number).

@egc112 The SSH drop of connections is reproducible on every click of "Save & Apply", which triggers a PBR restart. It also happens with other services like a PostgreSQL connection via `psql` command

PR #54, opened 2025-11-11
refactor pbr.user.dnsprefetch

betonmischer86:
Fixes [#27659](https://github.com/openwrt/packages/issues/27659) along with a general restructuring of the script.

PR #55, opened 2025-11-16
pbr: simplify resolver prefill helper to avoid blocking

pesa1234:
In my situation pbr fail to stop and fail to get service pbr status, when I enable /usr/share/pbr/pbr.user.dnsprefetch.

Openwrt also fail to backup...

After this commit if I enable the script all is back to normal.

- drop ubus listener and named pipes (mkfifo) from the dnsmasq resolver prefill helper to avoid potential deadlocks and blocking behaviour
- wait directly for pbr nft sets (nft list sets inet | grep pbr) with a bounded timeout before doing any lookups
- resolve domains by iterating over $packageDnsmasqFile and calling nslookup <domain> 127.0.0.1 in a simple loop
- only run the helper when resolverSetSupported is enabled, the $packageDnsmasqFile exists, and it actually contains at least one domain entry
- keep the helper running in a background subshell so pbr startup is not delayed while dnsmasq/policy domains are being resolved

Thanks

### DIFF
--- a/files/usr/share/pbr/pbr.user.dnsprefetch
+++ b/files/usr/share/pbr/pbr.user.dnsprefetch
@@ -4,10 +4,10 @@
 
 (
 	timeout_nft='10'
-	timeout_dnsmasq='20'
-	pipe_ubus="/tmp/pipe.ubus.$$"
-	pipe_nslookup="/tmp/pipe.nslookup.$$"
 	log_abort='domain names in policies not resolved'
+
+	entries=0
+	errors=0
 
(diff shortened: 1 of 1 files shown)


stangri:
@betonmischer86 could you please confirm the underlying issue has been addressed and this can be closed?

betonmischer86:
> @betonmischer86 could you please confirm the underlying issue has been addressed and this can be closed?

Yes, I never saw the issue repeat since the [workaround](https://github.com/mossdef-org/pbr/pull/54) was merged.

PR #87, opened 2026-03-11
pbr[1.2.2]: Add forwarding protection to restart

egc112:
Also add forwarding protection to restart function.


stangri:
Thank you Erik!

PR #96, opened 2026-04-22
pbr_get_gateway6: fix fallback producing invalid `ip -6 route` commands

brauliobo:
Problem
-------
pbr_get_gateway6() uses two strategies to discover the IPv6 next-hop for an interface:

  1. network_get_gateway6 (from /lib/functions/network.sh, parses the ubus `network.interface.<iface> status` object).
  2. A shell fallback that greps the interface's own addresses when (1) returns nothing usable ('', '::', '::/0', '::0/0').

The fallback (old line 274) was:

    gw="$(ip -6 a list dev "$dev" 2>/dev/null \
          | grep inet6 | grep 'scope global' | awk '{print $2}')"

Two independent defects:

 a) `awk '{print $2}'` runs over every matching line. On any interface
    with more than one global address — e.g. a DHCPv6 IA_NA plus SLAAC,
    an IA_PD assignment, or multiple delegated prefixes — the output is
    a newline-separated list, not a single address.
 b) `$2` is `<addr>/<prefixlen>` (from `ip -6 a` formatting). The `/<n>`
    suffix is not valid syntax for `ip route ... via`.

`gw6` is consumed by callers like (init.d/pbr:2247):

    try ip -6 route replace default via "$gw6" dev "$dev6" \
        table "$tid" metric "$uplink_interface6_metric"

With a multi-line, prefix-suffixed value quoted into `via`, `ip` sees one positional arg containing embedded newlines and returns an error.

Observable consequences on a real multi-WAN box
-----------------------------------------------
Reproducer environment: OpenWrt 25.12.2, three WANs — two of which receive multiple global IPv6 addresses from their ISPs:

    tim-wan:  2804:214:11:278d:62be:b4ff:fe1a:5cb5/64
              2804:214:11:278d::1/128
    str-wan:  fd0e:9b01:f8d7:10:62be:b4ff:fe1a:5cb7/64
              2803:9810:70fd:f610:62be:b4ff:fe1a:5cb7/64
              fd0e:9b01:f8d7:10::802/128
              2803:9810:70fd:f610::802/128

pbr logged the following every single run:

  ERROR: Command failed: ip -6 route replace default via \
    2804:214:11:278d:62be:b4ff:fe1a:5cb5/64 2804:214:11:278d::1/128 \
    dev tim-wan table 259 metric 128
  ERROR: Command failed: ip -6 route replace default via \
    fd0e:9b01:f8d7:10:62be:b4ff:fe1a:5cb7/64 \
    2803:9810:70fd:f610:62be:b4ff:fe1a:5cb7/64 \
    fd0e:9b01:f8d7:10::802/128 2803:9810:70fd:f610::802/128 \
    dev str-wan table 260 metric 128

Because the failure happens inside the main route-installation loop and netifd hotplug re-triggers pbr on any interface event (including the RA-refresh of the kernel-managed default route, which fires every router-lifetime period), pbr effectively enters a permanent reload loop: every ~60–75s it flushes and rebuilds its fw4 nftables chains. Each rebuild briefly perturbs conntrack mark save/restore and any in-flight UDP masquerade state, which on this box manifested as second-scale audio drops on WebRTC calls (Meet / Teams / Jitsi) across every WAN. pbr logs show the loop clearly:

  10:46:25  pbr ERROR ...
  10:47:02  pbr ERROR ...   (+37s)
  10:47:58  pbr ERROR ...   (+56s)
  10:49:04  pbr ERROR ...   (+66s)
  ...

Setting pbr.config.ipv6_enabled=0 silences the error loop and stops the forwarding disruption — confirming pbr's v6 fallback as the trigger.

Fix
---
Replace the fallback with a three-step cascade that returns a usable next-hop address (never a prefixed or multi-valued string):

 1. `ip -6 route show dev <dev>`: pick up the kernel's own default-route nexthop on that device (this is the address from the ISP's RA, typically an fe80::). This is what we actually want as `via`.
 2. `ip -6 neigh show dev <dev>`: first link-local neighbor flagged `router` — covers the case where step 1 missed because no default route has been planted in the main table for this device yet but an RA has been received.
 3. First global address on the device with `/<prefixlen>` stripped. Not ideal (it is our own address, not a real upstream gateway), but at least syntactically valid so the resulting `ip route` command does not crash pbr into a reload loop. `awk … exit` guarantees single-line output.

network_get_gateway6 remains the primary path — this only touches the fallback and does not change behaviour on interfaces where ubus already provides a nexthop.

Verification on the reproducer box after the patch -------------------------------------------------- pbr error count over the following 90s window: 0 (was 1 every ~60s). pbr "started with gateways:" log line, before vs after:

  before: wan_tim/tim-wan/192.168.1.1/
          2804:214:11:278d:62be:b4ff:fe1a:5cb5/64 2804:214:11:278d::1/128
          wan_str/str-wan/192.168.1.1/
          fd0e:9b01:f8d7:10:62be:b4ff:fe1a:5cb7/64
          2803:9810:70fd:f610:62be:b4ff:fe1a:5cb7/64
          fd0e:9b01:f8d7:10::802/128 2803:9810:70fd:f610::802/128

  after:  wan_tim/tim-wan/192.168.1.1/fe80::1
          wan_str/str-wan/192.168.1.1/fe80::7624:9fff:fed2:bc12
          wan_vivo/vivo-wan/192.168.15.1/fe80::c23d:d9ff:fe38:9e28

Each interface resolves to the upstream router's link-local address, which is the correct `via` value for `ip -6 route replace default`.

Note on the LuCI status display
-------------------------------
The "Service Gateways" field in LuCI previously showed GUA-looking values for misconfigured interfaces because the buggy fallback returned the interface's own global address. After the fix it shows the actual next-hop (fe80::...). That is a display change, not a regression — the old values were wrong, and attempting to use them as `via` was what caused the error loop in the first place.

### DIFF
--- a/files/etc/init.d/pbr
+++ b/files/etc/init.d/pbr
@@ -271,7 +271,15 @@ pbr_get_gateway6() {
 	is_uplink4 "$iface" && iface="$uplink_interface6"
 	network_get_gateway6 gw "$iface" true
 	if [ -z "$gw" ] || [ "$gw" = '::/0' ] || [ "$gw" = '::0/0' ] || [ "$gw" = '::' ]; then
-		gw="$(ip -6 a list dev "$dev" 2>/dev/null | grep inet6 | grep 'scope global' | awk '{print $2}')"
+		# Prefer the actual default-route nexthop on this device (usually fe80::).
+		gw="$(ip -6 route show dev "$dev" 2>/dev/null | awk '$1=="default" {print $3; exit}')"
+		# Fall back to a link-local neighbor advertised as router.
+		[ -z "$gw" ] && gw="$(ip -6 neigh show dev "$dev" 2>/dev/null | awk '/^fe80:.*router/ {print $1; exit}')"
+		# Last resort: first global address on the device, with prefix length stripped.
+		# The previous `awk '{print $2}'` (no line limit, prefix kept) produced a
+		# multi-line string like "2804::a/64\n2804::b/128" which turned `ip -6 route
+		# replace default via "$gw6" ...` into invalid syntax and failed every run.
+		[ -z "$gw" ] && gw="$(ip -6 a list dev "$dev" 2>/dev/null | awk '/inet6 .* scope global/ {sub(/\/.*/,"",$2); print $2; exit}')"
 	fi
 	eval "$1"='$gw'
 }


stangri:
Hello, @brauliobo !

Thank you for the elaborate PR/description. I appreciate you pointing out the runtime errors, I wonder:

> Reproducer environment: OpenWrt 25.12.2, three WANs — two of which receive multiple global IPv6 addresses from their ISPs

In this environment, what is option [uplink_interface6](https://docs.openwrt.melmac.ca/pbr/1.2.2/#uplink_interface6) is set to?

> Because the failure happens inside the main route-installation loop and netifd hotplug re-triggers pbr on any interface event (including the RA-refresh of the kernel-managed default route, which fires every router-lifetime period), pbr effectively enters a permanent reload loop: every ~60–75s it flushes and rebuilds its fw4 nftables chains.

On interface events the `pbr` should no longer rebuild the fw4 nftables chains, even in the dynamic routing tables mode it only modifies the routing tables.

There have been an issue of phantom IPv6 interface updates in OpenWrt for quite some time, that's why the `https-dns-proxy` for example can be set to ignore `wan6` interface updates. To minimize the problem for pbr, you may want to use netifd extensions with pbr 1.2.2 and 1.2.3.

PS. You seem to be a prime candidate to try mwan4 and pbr 1.2.3 ;)

egc112:
Thanks for reporting, I see your problem.
But it would be interesting to know how you arrived in this situation.
As far as I know PBR supports a number of VPN interfaces and exactly one WAN interface which is defined with `option uplink_interface / option uplink_interface6`.
(For multi wan you better use mwan3 or mwan4.)

So I would expect that PBR should only process one WAN interface and ignore the others.
But @stangri is the expert, I might be mistaken.

Furthermore the build in `network_get_gateway6` (from lib/functions/network.sh) should take care of getting the gateway for the netifd interfaces, not sure why that is not working for you, are those wan interfaces not netifd interfaces or is the `network_get_gateway6` flawed?

The fallback is for non netifd interfaces (e.g. openvpn which creates its own interface) but as those are point-to-point links the gateway is not used so that is why the problem has not surfaced.

But you are correct that the fallback code to get a gateway needs attention.

For our insight it would be helpful if you can share the following:
```
service pbr support
ip address show
```

stangri:
@brauliobo please review copilot's feedback, I'd like to commit it to this PR and merge both fixes.

files/etc/init.d/pbr
```diff
@@ -271,7 +271,15 @@ pbr_get_gateway6() {
...
-		gw="$(ip -6 a list dev "$dev" 2>/dev/null | grep inet6 | grep 'scope global' | awk '{print $2}')"
+		# Prefer the actual default-route nexthop on this device (usually fe80::).
+		gw="$(ip -6 route show dev "$dev" 2>/dev/null | awk '$1=="default" {print $3; exit}')"
+		# Fall back to a link-local neighbor advertised as router.
+		[ -z "$gw" ] && gw="$(ip -6 neigh show dev "$dev" 2>/dev/null | awk '/^fe80:.*router/ {print $1; exit}')"
+		# Last resort: first global address on the device, with prefix length stripped.
+		# The previous `awk '{print $2}'` (no line limit, prefix kept) produced a
+		# multi-line string like "2804::a/64\n2804::b/128" which turned `ip -6 route
+		# replace default via "$gw6" ...` into invalid syntax and failed every run.
+		[ -z "$gw" ] && gw="$(ip -6 a list dev "$dev" 2>/dev/null | awk '/inet6 .* scope global/ {sub(/\/.*/,"",$2); print $2; exit}')"
```
egc112, line 282: I also have my doubts about the necessity of this line.
Yes it will get a valid address but is that the gateway?
Many VPN interfaces are a point-to-point link which do not have/need a gateway

egc112:
I have looked some more into the reported problem and the proposed solution by the OP.

This is my take on the subject.

The `pbr_get_gateway4` and  `pbr_get_gateway6` are used to get the gateway for setting up routes for a supported interface.
Most of the default supported interfaces do not have/need a gateway as they are point-to-point links, but e.g. the wan/wan6 interfaces are broadcast linked interfaces which do need a gateway for the proper routing.
The netifd interfaces (e.g. wan/wan6) can be queried with the build in network_get_gatewayX which reads the nexthop from `ubus call "network.interface.${iface}" status | jsonfilter -e "@.route[0].nexthop"`, but there are interfaces which do not have that set or are not netifd interfaces.
So a fallback has to be added to get the gateway for those interfaces.
The present fallback is not suitable for this as the OP demonstrated.
A possible better solution could be e.g.
```
gw="$(ip -4 route show dev "$dev" table all 2>/dev/null | awk '$1=="default" {for (i=1; i<NF; i++) if ($i=="via") {print $(i+1); exit}}')"
```
It is possible that the interfaces do not have a default route so a fallback to deal with this can be
```
gw="$(ip -4 route get 1.1.1.1 dev "$dev" 2>/dev/null | awk '/via/ {print $3; exit}')"
```
This does not work for IPv6 as source routing is used but for IPv6 can be used as fallback:
```
[ -z "$gw" ] && gw="$(ip -6 neigh show dev "$dev" 2>/dev/null | awk '/^fe80:.*router/ {print $1; exit}')"
```

If that still does not provide a gateway and the interface does not have a point-to-point link an error should be raised

This means that for a point-to-point interface there is no gateway and of course it is not needed either.

Therefore the code dealing with the routing :
```
https://github.com/mossdef-org/pbr/blob/03915fc8cb8d050de10eb2c9731d0025cb1b8360/files/etc/init.d/pbr#L2283
```
Needs to be adapted for dealing with no gateway for point-to-point links and if no strict enforcement is used.

This should also  address the problem of the OP regarding the wrong parsing of the gateway.

egc112:
@brauliobo  your ideas have been used and are incorporated in 1.2.2-r15.
See: https://forum.openwrt.org/t/policy-based-routing-pbr-package-discussion/140639/2787?u=egc

Please test and let us know

egc112:
I trust the testing was satisfactory if not please raise an issue or reopen this PR

Thanks for your contribution

PR #99, opened 2026-05-01
Add gateway/gateway6 options to policies

yannleretaille:
I've been using pbr for a while in conjunction with a multi-peer wireguard interface. Currently, pbr does not allow setting a gateway/next-hop ip for a interface and instead only uses the auto-derived default gateway. I initially manually replaced the routes to make things work, but this is very brittle and doesn't allow having multiple policies routing to different peers on the same wireguard interface. 

This PR adds optional gateway/gateway6 overrides to individual policies and creates individual pbr profiles and tables for the base interface and each distinct gateway/gateway6 overide combination by adding a suffix. For example:
* **pbr_wg:** wg + default route (auto)
* **pbr_wg_gw_10_100_1_1:** wg + gateway
* **pbr_wg_cybr_gw6_2001_db8_2:** wg + gateway6
* **pbr_wg_cybr_gw_10_100_1_1_gw6_2001_db8_2:** wg + gateway + gateway6

Most of the code is a few added tests to ensure the pairs are profile/table/mark creation and rule assignment work correctly for all possible combinations. This has been working well with on my router (25.12.0) and I additionally tested the different cases on an OpenWrt x86_64 VM. 

I'm not sure how frequent this use case is, but I would assume it's become a lot more relevant with the rise in popularity of wireguard. Let me know if I should remove the "Example Custom Next-Hop" policy.

I've created a corresponding PR to luci-app-pbr here: [luci-app-pbr/#21](https://github.com/mossdef-org/luci-app-pbr/pull/21)

AI use disclaimer: I used codex to generate most of the code based on my specification, which was reviewed and manually tested by me.

### DIFF
--- a/files/etc/config/pbr
+++ b/files/etc/config/pbr
@@ -58,3 +58,11 @@ config policy
 	option interface 'wan'
 	option dest_addr 'plex.tv my.plexapp.com emby.media app.emby.media tv.emby.media'
 	option enabled '0'
+
+config policy
+	option name 'Example Custom Next-Hop'
+	option interface 'wan'
+	option gateway '192.0.2.1'
+	option gateway6 '2001:db8::1'
+	option dest_addr 'example.com'
+	option enabled '0'
--- a/files/etc/init.d/pbr
+++ b/files/etc/init.d/pbr
@@ -230,6 +230,8 @@ load_validate_policy() {
 	local src_port
 	local dest_addr
 	local dest_port
+	local gateway
+	local gateway6
 	uci_load_validate "$packageName" 'policy' "$1" "${2}${3:+ ${3}}" \
 		'name:string:Untitled' \
 		'enabled:bool:1' \
@@ -240,6 +242,8 @@ load_validate_policy() {
 		'src_port:list(neg(or(portrange,string)))' \
 		'dest_addr:list(neg(or(host,network,string)))' \
 		'dest_port:list(neg(or(portrange,string)))' \
+		'gateway:or("", ip4addr)' \
+		'gateway6:or("", ip6addr)' \
 	;
 }
 
--- a/files/lib/pbr/config.uc
+++ b/files/lib/pbr/config.uc
@@ -86,6 +86,8 @@ function create_config(uci_mod, ubus_mod, pkg) {
 		dest_port: ['string', ''],
 		proto:     ['string', ''],
 		chain:     ['string', ''],
+		gateway:   ['string', ''],
+		gateway6:  ['string', ''],
 	};
 
 	const dns_policy_schema = {
--- a/files/lib/pbr/pbr.uc
+++ b/files/lib/pbr/pbr.uc
@@ -46,6 +46,7 @@ function create_pbr(fs_mod, uci_mod, ubus_mod) {
 
 	let iface_registry = {};
 	let iface_priority = '';
+	let iface_mark_next = '';
 	let ifaces_triggers = '';
 	let service_start_trigger = '';
 	let process_dns_policy_error = false;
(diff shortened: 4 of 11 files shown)


egc112:
Maybe I misunderstand what you want but WireGuard is a point-to-point link and does not have/need a gateway.

WireGuard itself is using cryptokey routing for internally routing to peers, see:
https://www.wireguard.com/

By defining the addresses of the peers and the Allowed IPs and  enabling route allowed IPs the kernel routing is also setup to route to the Allowed IPs. That way it is  possible to route to multiple peers.
I have a WireGuard interfaces which has 4 peers all are other routers with their own subnets and as the subnets are set as Allowed IPs and routing of allowed IPs is enabled there is kernel routing available to all the other peers and no need for any Policy based routing.

Furthermore If you want specific traffic to go to a specific peer then just create an extra WireGuard interface with just that peer, you can have as many WireGuard interfaces as you want.

Perhaps you can post in the forum what your use case is and what you want to see how we can solve that?

yannleretaille:
Thank you for getting back to me this quickly!

> Furthermore If you want specific traffic to go to a specific peer then just create an extra WireGuard interface with just that peer, you can have as many WireGuard interfaces as you want.

That's exactly my use case. I have a wireguard network of multiple routers and nodes and use pbr to have certain traffic go through certain peers. Creating separate interfaces for every peer would make things quite unwieldy and harder to manage, hence this PR.

I wonder if maybe gateway is not quite the right term. I initially called the option "via" as used in linux routing, and also considered "nexthop". Eventually I settled on gateway as thats the term used throughout uci in OpenWrt.

egc112:
> Creating separate interfaces for every peer would make things quite unwieldy and harder to manage, hence this PR.

But for this solution you do no need to add extra code which needs maintenance, you do not need to add an extra interface element which will confuse users and you do not need to mess with gateways.

If you administer/deploy a large number of nodes have a look into using [uci scripts](https://openwrt.org/docs/guide-user/services/vpn/wireguard/automated) or programs like [unetd](https://openwrt.org/docs/techref/unetd) netmaker or headscale

But still kudos for your work and thanks for sharing.

yannleretaille:
Thank you again for taking the time to consider this.

> do not need to add an extra interface element which will confuse users 
I'm not sure if I agree here. I personally was confused to not find this extra option and ended up going through the documentation several times and looking for hidden uci options. So for me the opposite was the case and the lack of the option was the confusing part. It would have been fine if it'd just been an uci option not exposed in luci though.
 
I'm used to having the option to define a gateway when adding routes and this is also how OpenWrts static route configuration is modeled. PBR also manages routes/tables, but doesn't have the option.

> If you administer/deploy a large number of nodes have a look into using [uci scripts](https://openwrt.org/docs/guide-user/services/vpn/wireguard/automated) or programs like [unetd](https://openwrt.org/docs/techref/unetd) netmaker or headscale

I agree these tools can help with managing and provisioning meshes, but that doesn't really help with the issue I'm trying to address here in PBR. AFAIK, even with something like netmaker,  I would still need a gateway option in PBR if I'd like to route certain traffic through certain nodes.

Without going into too much detail, the WireGuard mesh and PBR in my setup are used in the context of censorship circumvention, for which PBR has been very helpful. It also means that I don't want to use something like tailscale that needs a centralized control plane (and also is a commercial project).   

As I elaborated before, rewriting the routes in PBR-generated tables with external scripts is what I did before. It's very brittle and really difficult when trying to do multiple routes/tables per interface.

I’m using WireGuard as the practical example because that is my setup, but the same need arises for any interface with multiple valid next hops. Some homelabers might have transit VLANs with multiple upstream routers/firewalls/etc., or WANs with more than one provider gateway, or other overlay/tunnel interfaces with multiple reachable routers.

This PR adds ~100 lines to pbr.uc and I personally find the profile naming logic and route setup fairly easy to follow. I wanted to make sure the tests are exhaustive and cover all relevant gateway/gateway6 combinations, but I'd be happy to look into simplifying/condensing them if you think that would ease the maintenance burden.

egc112:
Besides my earlier objectives, I see a more fundamental problem, because of the cryptokey routing you have to align the allowed ips which might result in overlap which can lead to undetermined behaviour, alternatively you have to force the peer side to masquerading in which case you loose logging and access control for bidirectional traffic (site-to-site)

All of which you do not have if you just create an interface and if you do that you can do away with your brittle routing. :)

So I am sorry but at this.moment  I am not convinced but I will discuss it with the maintainer.

egc112:
We have discussed your PR and as there already are ways to accomplish what you want, although in a different way then what you would like, we do not see enough value in implementing it at the moment.

For large site-to-site setups (Server with multiple peers) you can use the cryptokey routing with routing of the allowed IP's in which case you do not need to use any PBR (For an example see [WireGuard Client Setup Guide](https://raw.githubusercontent.com/egc112/OpenWRT-egc-add-on/main/notes/OpenWRT%20WireGuard%20Client%20Setup%20guide%20using%20LuCi.pdf)
chapters about `site-to-site setup` and `multi-site setup`).

For more specific routing you can create an extra interface/tunnel.

But we thank you for your contribution.

PR #102, opened 2026-05-10
pbr-1.2.2: flush rt-tables on stop and small optimizations

egc112:
On stop the routing tables are not flushed.
This patch should correct that

Also some small optimisations for querying if an interface is point-to-point

minor things so no hurry in implementing

Please let @copilot do its thing to check


stangri:
Merged with thanks Erik!

PR #104, opened 2026-05-11
pbr[1.2.3]: emit pbr_mark chain and fwmark rule for ifaces with manual ip4table

Per-Terra:
## Summary

In `interface_routing.create`, netifd-managed interfaces share an unconditional early return with mwan4. The early return skips installation of the `pbr_mark_${mark}` nft chain and the `ip rule fwmark ${mark}/${fw_mask}` rule — but policy processing still emits `goto pbr_mark_${mark}`. On my hardware the resulting atomic nft ruleset fails the dry-check, pbr refuses to install it, and policies targeting a netifd-managed uplink silently don't take effect.

This PR splits the early return so:

- **mwan4** keeps its current behavior — its mark chain comes from `env.mwan4_interface_chain[iface]` and is set up elsewhere.
- **netifd** now installs the mark chain via the existing `nft.ensure_mark_chain` helper, plus an `ip rule fwmark ${mark}/${fw_mask}` pointing at the netifd-managed table read from `network.<iface>.ip4table` / `ip6table`. The default route inside that table is owned by netifd and is left untouched.

Possibly related to the symptoms reported in #92.

## Background and disclaimer

I'm not deeply familiar with pbr internals or OpenWrt's netifd path — I came at this from the user side after noticing that policies stopped applying once my uplink was set up via netifd. What I'm proposing here is the same fix I'd been running locally as a shell hot-patch against the installed `/etc/init.d/pbr` on my router to survive reboots. Per `CONTRIBUTING.md` I rebased that change onto the latest version branch (`1.2.3`), where the same logic now lives in ucode.

Given that I'm new to this code, I'd really appreciate a careful review — there are likely subtleties in the netifd path I'm missing.

## Scope

The original shell hot-patch only touched the `create` action, which is what I've actually exercised on my router. I've intentionally kept this PR's scope identical:

- `destroy` and `reload` for netifd ifaces are **not** modified in this PR.
- On service stop, the nft table teardown removes the mark chain. The leaked `ip rule` survives, but its fwmark selector can no longer match (nothing sets that mark anymore) so it should be harmless.
- `ip rule replace` is idempotent, so repeated service starts don't accumulate rules.
- Single-iface reload (`on_interface_reload`) for a netifd iface is a corner case I haven't exercised on hardware.

If you'd prefer the same patch symmetrically applied to `destroy`/`reload`, I'm happy to extend it — I just didn't want to ship code I haven't actually tested.

## Reused helpers (no new code paths)

[...]

PR #105, opened 2026-05-12
pbr[1.2.3]: tighten is_netifd_interface() to require env.netifd_mark

Per-Terra:
Re-take of #104 along the direction @stangri proposed in https://github.com/mossdef-org/pbr/pull/104#issuecomment-4430726271 — instead of adding a netifd-specific branch to `interface_routing.create`, tighten `is_netifd_interface()` itself.

`is_netifd_interface()` previously classified by UCI alone (any iface with `ip4table` or `ip6table` set), which over-included user-manual two-uplink setups that pbr did not put in place. In dynamic-routing-tables mode this caused `interface_routing 'create'` to early-return for such ifaces, leaving the per-mark chain and `ip rule fwmark` uninstalled while policy processing still emitted `goto pbr_mark_${mark}` — the resulting ruleset failed `nft -c -f` and pbr aborted on start.

The helper is now narrowed to ifaces present in `env.netifd_mark`, populated from `pkg.nft_netifd_file` — i.e. ifaces pbr itself set up via its netifd extensions. The check now mirrors `is_mwan4_interface()`'s `env.mwan4_mark[iface]`.

One subtle side effect: user-manual `ip4table` ifaces are now added to `ifaces_triggers`, so pbr reloads on their dev up/down — the intended behaviour once pbr does manage them.

### DIFF
--- a/files/lib/pbr/network.uc
+++ b/files/lib/pbr/network.uc
@@ -132,10 +132,7 @@ function create_network(fs_mod, config, sh, pkg, platform, V) {
 	function is_ignore_target(iface) { return lc(iface) == 'ignore'; }
 	function is_netifd_table(name) { let c = readfile('/etc/config/network') || ''; return index(c, name) >= 0 && !!match(c, regexp('ip.table.*' + name)); }
 	function is_netifd_interface(iface) {
-		let ctx = config.uci_ctx('network');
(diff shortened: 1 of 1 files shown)


stangri:
@Per-Terra apologies for the delay in merging. This is a very elegant fix, kudos!

PR #114, opened 2026-05-26
pbr-1.2.2: guard pbr_get_gateway6 when ipv6 is disabled

egc112:
Even if IPv6 is disabled pbr_get_gateway6 is called. 
Disable this by guarding with ipv6_enabled check 
This can prevent an error when a broadcast interface has no IPv6.

### DIFF
--- a/files/etc/init.d/pbr
+++ b/files/etc/init.d/pbr
@@ -273,6 +273,7 @@ pbr_get_gateway4() {
 	eval "$1"='$gw'
 }
 pbr_get_gateway6() {
+	[ -z "$ipv6_enabled" ] && return 0
 	local iface="$2" dev="$3" gw
 	is_uplink4 "$iface" && iface="$uplink_interface6"
 	network_get_gateway6 gw "$iface" true


stangri:
> Even if IPv6 is disabled pbr_get_gateway6 is called. Disable this by guarding with ipv6_enabled check This can prevent an error when a broadcast interface has no IPv6.

Let's see what copilot says, but I think it may have been intentional so we could discover if IPv6 on the system is enabled even if ipv6 support for pbr is off.

egc112:
Yes could be but this could be triggering the warning you are seeing.

You can try and add it to your router and see if the warning is gone.

I think what happens is that when an interface has no IPv6 the pbr_get_gateway6 is run anyway.
The interface does not have functioning ipv6, so no ipv6 gateway and if it is not a point-to-point interface the warning is triggered.

stangri:
Thanks, apparently pbr_get_gateway6 is not used to detect IPv6, merging with gratitude and I'll pick this for 1.2.3!

egc112:
Let me know if the warning you see is gone , I have high hopes 🤞

PR #115, opened 2026-05-26
pbr-1.2.2: correct multiple dns source entries

egc112:
Thanks to @AndrewZ : https://forum.openwrt.org/t/policy-based-routing-pbr-package-discussion/140639/2801?u=egc

Multiple MAC entries in DNS source rule were generating an error because the whole line was evaluated

Correct by evaluating each entry in the source

### DIFF
--- a/files/etc/init.d/pbr
+++ b/files/etc/init.d/pbr
@@ -2115,15 +2115,17 @@ dns_policy_process() {
 	dest_dns_ipv4="$(str_first_value_ipv4 "$dest_dns")"
 	dest_dns_ipv6="$(str_first_value_ipv6 "$dest_dns")"
 	if is_supported_interface "$dest_dns_interface"; then
-		local d
+		local d s
 		for d in $(uci -q get network."$dest_dns_interface".dns); do
-			if ! is_family_mismatch "$src_addr" "$d"; then
-				if is_ipv4 "$d"; then
-					dest_dns_ipv4="${dest_dns_ipv4:-${d}}"
-				elif is_ipv6 "$d"; then
-					dest_dns_ipv6="${dest_dns_ipv6:-${d}}"
+			for s in $src_addr; do
+				if ! is_family_mismatch "$s" "$d"; then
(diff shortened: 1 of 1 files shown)


stangri:
@egc112 Claude provided another insight:

```
  The cleaner refactor would be to move the v4/v6 server resolution into the filter_options loop — inside each group iteration, you already know the family (the group name tells you), and
  filtered_value_src_addr is single-family by construction, so the family check becomes implicit. Roughly:

  for filter_group_src_addr in $filter_list_src_addr; do
      filtered_value_src_addr="$(filter_options "$filter_group_src_addr" "$src_addr")"
      [ -n "$src_addr" ] && [ -n "$filtered_value_src_addr" ] || continue

      # Resolve dest_dns server by the group's family
      case "$filter_group_src_addr" in
          *ipv4*) [ -z "$dest_dns_ipv4" ] && continue ;;
          *ipv6*) [ -z "$dest_dns_ipv6" ] && continue ;;
      esac

      dns_policy_routing "$name" "$filtered_value_src_addr" "$dest_dns" "$uid" \
          "$dest_dns_port" "$dest_dns_ipv4" "$dest_dns_ipv6"
  done

  …with the upstream dest_dns_ipv4/dest_dns_ipv6 pre-resolution happening either (a) just from the user-set dest_dns literals (no family check needed — they're already filtered by
  is_ipv4/is_ipv6), or (b) by iterating each DNS server in the interface and recording both v4 and v6 candidates without filtering by src_addr family at that stage (defer the family
  decision to the group loop below).

  Same bug exists in 1.2.3's ucode at files/lib/pbr/pbr.uc:566 — V.is_family_mismatch(src_addr, d) with multi-value src_addr. PR #115 will need to be forward-ported to 1.2.3, either as the
  same per-entry loop or as the deeper refactor.
```

Should I ask it to fix inside filter_options?

egc112:
That is fine by me anything which works works 😄 

Will test it tomorrow

PR #134, opened 2026-07-20
pbr-1.2.3: refactor restart()

egc112:
`restart_service` is not working at all, however when doing a `service pbr restart` automatically a `stop` and `start` is done which does indeed trigger a restart. 
But this will cause leakage as forwarding is not stopped  on `stop` We need a separate `restart ()` just as in 1.2.2 with proper stopping of forwarding at the start of the `restart ()` function to block leakage of traffic


stangri:
Thank you @egc112!

PR #139, opened 2026-07-28
pbr-1.2.3: port xfrm/point-to-point support

egc112:
Add is_xfrm(), is_point_to_point(), is_p2p() device-level detectors and an is_xfrm_interface() protocol-level detector to network.uc, porting the shell version's v1.2.2 is_p2p patch.

- is_xfrm_interface(iface): true when the interface's proto starts with "xfrm"; wired into is_tunnel() alongside the existing tunnel protocol checks.
- is_point_to_point(dev)/is_xfrm(dev): parse `ip -o link show` / `ip -o -d link show` output for a given device. Kept as a literal match of the bash reference: is_point_to_point() uses a plain substring check (bash uses `grep -q`, no -w), is_xfrm() uses a word-boundary regex (bash uses `grep -qw`) to avoid false-matching "xfrm" inside an unrelated device name.
- is_p2p(dev) = is_point_to_point(dev) || is_xfrm(dev).

Fix get_gateway4()/get_gateway6() to use is_p2p(dev) instead of grepping `ip address show` output for POINTOPOINT, which doesn't reliably report the flag for all interface types and never detected xfrm devices at all.

Replace all 8 point-to-point fallback checks in pbr.uc's interface_routing() create/reload paths (ipv4+ipv6) with net.is_p2p(dev4)/net.is_p2p(dev6), removing the same `ip address show`-based POINTOPOINT grep.

Verified against a locally built ucode interpreter: both files compile clean as modules, and is_point_to_point/is_xfrm/is_p2p were runtime-tested against real `ip link show` output for tun, tap, and xfrm devices, plus get_gateway4()'s warning-suppression path.

### DIFF
--- a/files/lib/pbr/network.uc
+++ b/files/lib/pbr/network.uc
@@ -110,10 +110,31 @@ function create_network(fs_mod, config, sh, pkg, platform, V) {
 		return i;
 	}
 	function is_xray(iface) { return get_xray_traffic_port(iface) != null; }
+	function is_xfrm_interface(iface) { let _p = network_get_protocol(iface); return _p != null && substr(_p, 0, 4) == 'xfrm'; }
 	function is_tunnel(iface) {
 		return is_dslite(iface) || is_l2tp(iface) || is_oc(iface) || is_ovpn(iface) ||
 			is_pptp(iface) || is_softether(iface) || is_netbird(iface) ||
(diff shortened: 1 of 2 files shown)


egc112:
Will bump version and invite testers

PR #140, opened 2026-07-29
pbr-1.2.2: add is_xfrm_interface, use in is_tunnel; is_xfrm/is_point_…

egc112:
…to_point for is_p2p

Introduce is_point_to_point() and is_xfrm() as focused device-level checks:

  is_point_to_point() - detects POINTOPOINT via `ip -o link show`
  is_xfrm()             - detects xfrm interfaces via `ip -o -d link
                           show`, matching on the "xfrm" type marker
                           rather than the interface name, so it isn't
                           fooled by renamed devices

Fold both into a single is_p2p() and use it everywhere the old inline `ip -o link show ... | grep POINTOPOINT` check was duplicated: the gateway-warning guards in pbr_get_gateway4/6, and the eight IPv4/IPv6 route-replace call sites. This also fixes a pre-existing inconsistency where one of those ten call sites used `ip address show` instead of `ip -o link show`; it happened to still work but was unintentional drift from the other nine. Gateway-warning suppression for xfrm interfaces is now applied uniformly alongside point-to-point.

Separately, add is_xfrm_interface() and wire it into is_tunnel() so xfrm interfaces (e.g. IPsec) are recognized as supported/tunnel interfaces, consistent with the existing VPN protocol checks (WireGuard, OpenVPN, NetBird, etc). Unlike is_xfrm(), which expects a resolved Linux device name (used by is_p2p's device-based callers), is_tunnel is always called with a logical UCI interface name, so is_xfrm_interface checks the "xfrm" proto via network_get_protocol instead, matching the pattern already used by is_dslite/is_l2tp/is_oc/ is_pptp.

### DIFF
--- a/files/etc/init.d/pbr
+++ b/files/etc/init.d/pbr
@@ -268,8 +268,8 @@ pbr_get_gateway4() {
 		# Fall back to ip route get using "table all" might already work
 		[ -z "$gw" ] && gw="$(ip -4 route get 1.1.1.1 oif "$dev" 2>/dev/null | awk '/via/ {print $3; exit}')"
 		# Raise warning if no gw and not point-to-point
-		{ [ -z "$gw" ] && ! ip address show dev "$dev" 2>/dev/null | grep -q "POINTOPOINT"; } && output_warning "Unknown IPv4 Gateway for interface:'$iface' device:'$dev'"
-		#{ [ -z "$gw" ] && ! ip address show dev "$dev" 2>/dev/null | grep -q "POINTOPOINT"; } && json add warning 'warningInterfaceRoutingUnknownGateway' "IPv4 interface:[$iface]; device:[$dev]"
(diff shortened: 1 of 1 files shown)


egc112:
Asked co-pilot:

> All tests and smoke checks passed on target hardware:
> 
>     shellcheck passed.
>     is_point_to_point, is_p2p and is_xfrm behaved as expected for eth0 and wg_mullv_us (wg_mullv_us is detected as point-to-point).
>     ip route/neighbor outputs match expectations.
> 
> Behavioral notes:
> 
>     xfrm interfaces are intentionally treated as p2p and the JSON warning format change matches 1.2.3.
>     No blocking issues remain; this is ready to merge.
> 
> Optional, non-blocking suggestions: consider moving the new helper definitions earlier in the script (defensive), and adding a small CI smoke-test to catch environment differences.

PR #142, opened 2026-08-02
pbr-1.2.3: filter_options: fix negation state leaking across tokens

egc112:
filter_options() mutated its own `opt` parameter in place the first time it encountered a negated ('!'-prefixed) token:

    if (str_contains(opt, '_negative')) {
        if (substr('' + v, 0, 1) != '!') continue;
        opt = replace(opt, '_negative', '');
    }

Since `opt` is reassigned inside the loop, this permanently strips '_negative' off for every remaining iteration, silently disabling the "must start with !" requirement for the rest of that call's token list. On top of that, the non-negative branch never checked for the *absence* of '!' either, so a plain call like filter_options('ipv4', ...) matched negated tokens too.

Net effect: a single token could be classified into both the positive and negative group for its address family whenever a policy's src_addr/dest_addr mixed a negated entry with anything else of the same family (or even a single negated entry on its own, since the positive branch matched it unconditionally). Callers that loop over both groups (dns_policy_process, policy_process) would then route the same address(es) more than once, producing duplicate/overlapping nft rules - amplified further in policy_process's nested src x dest loop (up to 4x for a single negated src + single negated dest).

Fix: compute is_negative/base_opt once, before the loop, and require each token's negation state to match the group being filtered rather than mutating shared state as we go. Each token now lands in exactly one of the ten filter groups.

Verified via simulation that every token previously matched under any group is still matched exactly once post-fix, so the "unknown entry" validation in policy_process is unaffected.

### DIFF
--- a/files/lib/pbr/validators.uc
+++ b/files/lib/pbr/validators.uc
@@ -57,16 +57,17 @@ function is_family_mismatch(a, b) {
 
 function filter_options(opt, values) {
 	if (!values) return '';
+	let is_negative = str_contains(opt, '_negative');
+	let base_opt = is_negative ? replace(opt, '_negative', '') : opt;
 	let parts = split(trim('' + values), /\s+/);
 	let ret = [];
 	for (let v in parts) {
-		if (str_contains(opt, '_negative')) {
-			if (substr('' + v, 0, 1) != '!') continue;
-			opt = replace(opt, '_negative', '');
-		}
-		let check_val = replace(v, '!', '');
+		let negated = (substr('' + v, 0, 1) == '!');
+		if (is_negative && !negated) continue;
+		if (!is_negative && negated) continue;
+		let check_val = negated ? substr('' + v, 1) : '' + v;
 		let ok = false;
-		switch (opt) {
+		switch (base_opt) {
 		case 'phys_dev': ok = is_phys_dev(check_val); break;
 		case 'mac_address': ok = is_mac_address(check_val); break;
 		case 'domain': ok = is_domain(check_val); break;


egc112:
Co-pilot review:
Recommendation: Merge — the force-push fixes the negation-state leak and is low-risk.

I re-reviewed the updated diff (files/lib/pbr/validators.uc). The change now:

    computes is_negative/base_opt once,
    enforces per-token negation matching,
    only strips a leading '!' (substr(...,1)),
    switches on base_opt so tokens are classified into exactly one group.

This directly addresses the bug and my prior blocking suggestion, so it's safe to push. Optional follow-ups you can do later (not required to merge):

    add a small regression test that verifies mixed negated/non-negated tokens produce no duplicates,
    compute base_opt via an explicit suffix check + slice (rather than a global replace) and mark precomputed vars as constants for clarity.

stangri:
Late to comment, do I understand correctly that after this, a single ! Is in effect until the end of the entries/family of the policy? So !a b c = !a !b !c?

egc112:
The whole negation needs a careful look at also because PBR works with sets and you cannot have a negation in a set.

config policy
	option name 'wan'
	option interface 'wan'
	option src_addr ' !192.168.9.4/32 192.168.19.0/24 '

Old situation
    chain pbr_prerouting {
            counter packets 4515 bytes 680732 meta mark & 0x00ff0000 != 0x00000000 return
            ip saddr != { 192.168.9.4, 192.168.19.0/24 } counter packets 3919 bytes 591767 goto pbr_mark_0x010000 comment "wan"
    }

Everything is negated not what is the intention of this policy is.

New situation:
        chain pbr_prerouting {
                counter packets 18 bytes 1789 meta mark & 0x00ff0000 != 0x00000000 return
                ip saddr 192.168.19.0/24 counter packets 0 bytes 0 goto pbr_mark_0x010000 comment "wan"
                ip saddr != 192.168.9.4 counter packets 16 bytes 1437 goto pbr_mark_0x010000 comment "wan"
        }

This looks better, i think

But as said it needs a careful look at. 
Something for you when you have more time ? 😉

PR #144, opened 2026-08-03
tests: add POSIX shell test for DNS port string generation

egc112:
Add unit test for PR https://github.com/mossdef-org/pbr/pull/141

PR #147, opened 2026-08-10
pbr-1.2.3: dns_policy_routing bracket IPv6 dnat target when port is set

egc112:
`dns_policy_routing` built the IPv6 dnat target by appending `':' + dest_dns_port` directly to the address:

    dnat ip6 to 2606:4700:4700::1111:5353

nft requires an IPv6 address to be bracketed when a port follows it, since an unbracketed address is ambiguous — see nft(8):

    ip6 nat prerouting tcp dport 2222 dnat to [1ce::d0]:22

The rule is rejected and loading of `/var/run/pbr.nft` aborts, so any DNS policy pointing at an IPv6 resolver on a non-default port failed to apply.

This emits `[<addr>]:<port>` instead. IPv4 targets are unchanged, and neither family is bracketed when `dest_dns_port` is unset.

Same family of bug as #141, which fixed the trailing-colon case for the unset port but left the bracketing untouched.

## Testing

Extends `tests/04_policies/08_dns_policy_dest_port` (added in #144) with the IPv6-with-port case and a guard against brackets leaking into IPv4 or portless targets — 10 checks, up from 7.

Verified in both directions: full suite is 45/45 with the fix, and with `pbr.uc` reverted to the unfixed version the file scores 9/10, failing on exactly the new `v6_custom_bracketed` assertion. The seven pre-existing checks still hold on unfixed code, so the fold didn't change what #144 was testing.

### DIFF
--- a/files/lib/pbr/pbr.uc
+++ b/files/lib/pbr/pbr.uc
@@ -315,7 +315,12 @@ function create_pbr(fs_mod, uci_mod, ubus_mod) {
 			let inline_set_ipv4_empty = false, inline_set_ipv6_empty = false;
 	
 			let dest4 = 'dport 53 dnat ip to ' + dest_dns_ipv4 + (dest_dns_port ? ':' + dest_dns_port : '');
-			let dest6 = 'dport 53 dnat ip6 to ' + dest_dns_ipv6 + (dest_dns_port ? ':' + dest_dns_port : '');
+			// nft requires the IPv6 address to be bracketed when a port follows it.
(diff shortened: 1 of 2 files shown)


egc112:
Co-pilot comment:

> LGTM — good to merge.
> 
> This correctly brackets IPv6 dnat targets when a port is present, strips existing brackets to avoid duplication, and adds focused tests covering the regression and edge cases (bracketed input, scoped addresses, and IPv4/portless guards). Change is small, well-tested, and low-risk. Please merge once CI is green.

PR #149, opened 2026-08-12
pbr-1.2.3: guard IP rule priority exhaustion

egc112:
## Problem

`interface_enumerate()` allocates an IP rule priority per supported interface by counting **down** from `cfg.uplink_ip_rules_priority`, with no floor check. The parallel firewall mark allocation, which counts **up**, is already guarded by `errorInterfaceMarkOverflow` — the priority descent had no equivalent.

Once the priorities run out, pbr issues `ip rule add ... priority 0`, which collides with the kernel's own `from all lookup local` rule, and then continues into negative priorities. Same class of bug as 12b993766718.

## Fix

- **`pbr.uc` / `interface_enumerate()`** — skip the interface with `errorInterfacePriorityExhausted` once the counter would reach 0, mirroring the mark overflow guard directly above it. Skipped interfaces are never registered in `iface_registry`, and `interface_process.create()` already bails with `if (!existing) return 0;`, so they no-op downstream exactly like mark-overflow skips do today.
- **`pbr.uc` / `create_global_rules()`** — guard the matching descent there too. This is not just symmetry: `iface_priority` is shared leftover state and the `suppress_prefixlength` rule is issued unconditionally, so guarding only `interface_enumerate()` relocates the same bug one level up. The error is pushed at most once per call rather than once per WireGuard server.
- **`pkg.uc`** — add `errorInterfacePriorityExhausted`; bump `compat` to 35.
- **`init.d/pbr`** — update `initCompat` to 35.

## Severity

Low — hardening, not a live failure. `config.uc` clamps `uplink_ip_rules_priority` to 99..32765, so this only triggers with more than ~98 enabled and supported interfaces.

## Testing

New `tests/05_interfaces/12_priority_exhaustion_guard` drives 110 supported interfaces against `uplink_ip_rules_priority=99` and asserts 99 registered interfaces spanning priorities 99→1, zero non-positive priorities, and the expected exhaustion error count. Verified bug-sensitive: with the fix stashed it fails on `exhaustion_error_count`, `registered_interfaces` and `no_non_positive_priority`.

Full suite: 46/46 passing.

## Pairs with

The luci-app-pbr change setting `LuciCompat`/`rpcdCompat` to 35. Both must land together, or the version-mismatch warning fires.

### DIFF
--- a/files/etc/init.d/pbr
+++ b/files/etc/init.d/pbr
@@ -31,7 +31,7 @@ else
 fi
 
 readonly packageName='pbr'
-readonly initCompat='34'
+readonly initCompat='35'
 readonly _ucode="ucode -S -L /lib/${packageName} -L /lib/mwan4 /lib/${packageName}/cli.uc --"
 
 # shellcheck disable=SC1091
--- a/files/lib/pbr/pbr.uc
+++ b/files/lib/pbr/pbr.uc
@@ -929,6 +929,10 @@ function create_pbr(fs_mod, uci_mod, ubus_mod) {
 				push(state.errors, { code: 'errorInterfaceMarkOverflow', info: iface });
 				return;
 			}
+			if (+_iface_priority <= 0) {
+				push(state.errors, { code: 'errorInterfacePriorityExhausted', info: iface });
+				return;
+			}
 	
 			let dev4 = net.network_get_device(iface);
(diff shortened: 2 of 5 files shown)


egc112:
Co-pilot review:

> LGTM / Approve from a code-review perspective once CI is green and the luci-app-pbr compat bump is coordinated.

PR #150, opened 2026-08-13
pbr-1.2.3: restore the Tor policy warnings, split per option

egc112:
## What

A Tor policy silently discards several of the options a user can set on it, and one of them makes the emitted ruleset invalid. Nothing said so — the policy loaded without error and the rules looked plausible in `nft list`.

Tor is a dstnat redirect rather than a routed interface. `policy_routing()` forces the `dstnat` chain, clears `dest_port` and `proto`, and emits five rules hardcoding ports 53/80/443. So `dest_port`, `proto` and a non-`prerouting` `chain` are thrown away. `src_port` is worse than ignored: it survives into the rule, the now-empty `proto` is promoted to `tcp udp` because a port is set, and the result matches `tcp sport ... udp dport ...` — conflicting L4 protocols, which nft rejects.

This warns on all four. Warn and continue; nothing is rejected.

## Why they were missing

Warnings for exactly this were live between `f532ad3` (2022-12-14) and `d2aba93` (2024-02-16). That commit, *"better TOR support in nft"*, deleted the dedicated `policy_routing_tor_nft()`/`_iptables()` helpers and folded Tor into `policy_routing()`; the warning call sites lived inside those helpers and went with them. Confirmed unintentional rather than a deliberate removal.

The catalog strings were left behind in both repos and have been unreachable ever since — the WebUI has carried render entries for warnings that could never fire for over two years.

`warningTorUnsetParams` is **not** restored verbatim. It also told users to unset `src_addr`, which that same 2024 refactor turned into the correct way to match a Tor policy — so restoring it as-is would now be wrong advice. Split into `warningTorUnsetSrcPort` and `warningTorUnsetDestPort`, which also lets each carry its real consequence.

## Deliberately not warned about: `dest_addr`

A domain `dest_addr`, including `.onion`, is a supported Tor policy and works whenever DNS for that domain reaches Tor. pbr cannot observe whether it does — that depends on dnsmasq, which on OpenWrt declares `.onion` local through `/usr/share/dnsmasq/rfc6761.conf` and answers NXDOMAIN itself, overriding any `server=/onion/127.0.0.1#9053`. A warning there would fire exactly as loudly on a correct setup as on a broken one, so it is documented rather than warned about.

## Also

Drops the hardcoded `running_iptables: false` from the status payload. It has been unconditionally false since the iptables backend was removed in 1.1.7, and its only consumers were two dead branches in luci-app-pbr, removed in the companion PR.

## Testing

47/47 pass. New test `tests/04_policies/08_tor_policy_warnings` covers three Tor policies — one setting every discarded option, one setting only `src_addr`, one with an ordinary domain `dest_addr` — with 12 checks:

- each warning fires where it should, and none where it should not
- a domain `dest_addr` still builds its nftset and stays silent
- no policy is rejected
- Tor never reports an unknown gateway

That last check guards behaviour that is currently structural rather than asserted: Tor is registered by `interface_process.tor()` rather than the `foreach('network','interface')` loop, so it never reaches `get_gateway4()`/`get_gateway6()` and never warns about the device, gateway, table or mark it cannot have. Verified by mutation — injecting such a warning fails the test.

## ⚠️ Merge together with luci-app-pbr#39

These two are a matched pair. **mossdef-org/luci-app-pbr#39** carries the WebUI strings for the two new warning codes and the matching compat bump. Merging either alone leaves a compat mismatch — users would see the version-mismatch warning until both are in, and any warning fired by this PR would render as "Unknown warning".

## Follow-ups, not included here

- pbr reads only `/etc/tor/torrc`. OpenWrt's own Tor client guide writes its config to `/etc/tor/custom` and includes it via `tor.conf.tail_include`, so anyone following the official guide has DNSPort/TransPort values pbr never sees. It works today only because that guide uses 9053/9040 — the same numbers as pbr's fallback.
- The port parse accepts only the `address:port` form. A bare `DNSPort 9153` falls back to 9053 silently. Worth doing with the item above — widening the regex alone helps nobody whose config pbr is not reading.
- Only `udp dport 53` is redirected; there is no `tcp dport 53` rule, in 1.2.3 or 1.2.2. TCP DNS bypasses Tor entirely.
- In the Tor branch, `ipv4_error`/`ipv6_error` are tested inside the `tor_rules` loop and never reset, so one failed insertion is reported up to five times.
- `interface_process.tor('destroy')` is a no-op; the switch handles only create/reload/reload_interface.
- `tests/run_tests.sh` applies `sed '/get_rt_tables_id/,/^};/ s/m\[1\]/m[2]/'` to the patched `pbr.uc`. No line after that match starts with `};` at column 0, so the range runs to EOF and rewrites every `m[1]` in the tail, including both Tor port helpers — any test covering the Tor ports reads as broken (`:null`) even when the code is correct. This blocks tests for the two torrc items above, so it is worth fixing first.

### DIFF
--- a/files/etc/init.d/pbr
+++ b/files/etc/init.d/pbr
@@ -31,7 +31,7 @@ else
 fi
 
 readonly packageName='pbr'
-readonly initCompat='35'
+readonly initCompat='36'
 readonly _ucode="ucode -S -L /lib/${packageName} -L /lib/mwan4 /lib/${packageName}/cli.uc --"
 
 # shellcheck disable=SC1091
--- a/files/lib/pbr/pbr.uc
+++ b/files/lib/pbr/pbr.uc
@@ -630,6 +630,41 @@ function create_pbr(fs_mod, uci_mod, ubus_mod) {
 			push(state.errors, { code: 'errorPolicyNoInterface', info: name });
 			output.fail(); return 1;
 		}
+		if (net.is_tor(interface_name)) {
+			// Tor is a dstnat redirect, not a routed interface: the rules below
+			// hardcode ports 53/80/443 and force the dstnat chain, so these
+			// options are silently discarded (src_port additionally produces an
+			// invalid rule). Warn rather than reject -- a dest_addr holding
+			// ordinary resolvable domains is a legitimate Tor policy.
+			//
+			// History: these warnings were live from f532ad3 (2022-12-14) until
+			// d2aba93 (2024-02-16) 'better TOR support in nft', which deleted the
+			// dedicated policy_routing_tor_nft()/_iptables() helpers and folded
+			// Tor into policy_routing(). The call sites went with those helpers
+			// and were never re-added -- confirmed with the maintainer that this
+			// was unintentional rather than a deliberate removal. The catalog
+			// strings were left behind in pkg.uc and status.js, unreachable ever
+			// since. Deliberately not restored verbatim: the old
+			// warningTorUnsetParams also told users to unset src_addr, which that
+			// same refactor turned into the correct way to match a Tor policy, so
+			// its wording is now split per option.
+			//
+			// Deliberately NOT warned about: dest_addr, including '.onion'. A
+			// domain dest_addr is a supported Tor policy and works whenever DNS
+			// for the domain reaches Tor. pbr cannot see whether it does -- that
+			// depends on dnsmasq, which on OpenWrt declares '.onion' local via
+			// /usr/share/dnsmasq/rfc6761.conf -- so any such warning would fire
+			// just as loudly on a correct setup as on a broken one. Documented
+			// instead; see the Tor section of the README.
+			if (src_port)
+				push(state.warnings, { code: 'warningTorUnsetSrcPort', info: name });
+			if (dest_port)
+				push(state.warnings, { code: 'warningTorUnsetDestPort', info: name });
+			if (proto)
+				push(state.warnings, { code: 'warningTorUnsetProto', info: name });
+			if (chain && lc(chain) != 'prerouting')
+				push(state.warnings, { code: 'warningTorUnsetChainNft', info: name });
+		}
 		if (!net.is_supported_interface(interface_name) && !net.is_mwan4_strategy(interface_name)) {
 			push(state.errors, { code: 'errorPolicyUnknownInterface', info: name });
 			output.fail(); return 1;
@@ -2356,7 +2391,6 @@ function create_pbr(fs_mod, uci_mod, ubus_mod) {
 		result[name] = {
 			enabled: !!cfg.enabled,
 			running: platform.is_running_nft_file(),
-			running_iptables: false,
 			running_nft: nft.is_service_running_nft(),
 			running_nft_file: platform.is_running_nft_file(),
 			version: pkg.version,
--- a/files/lib/pbr/pkg.uc
+++ b/files/lib/pbr/pkg.uc
@@ -9,7 +9,7 @@
 const pkg = {
 	name: 'pbr',
 	version: 'dev-test',
-	compat: '35',
+	compat: '36',
 	config_file: '/etc/config/pbr',
 	debug_file: '/var/run/pbr.debug',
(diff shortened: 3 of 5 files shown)


egc112:
Thanks — went through all four. Three check out; on the helper extraction I'd push back.

### 1. Line range confirmed — `files/lib/pbr/pbr.uc:633-667`

35 lines, but that number is misleading: **25 are comment, 9 are code** — four `if` guards, four `push` calls, and the closing brace.

On extracting into a helper, I don't think the stated benefit applies here. The goal given is "easier to test in isolation", but nothing in this module is exported — `create_pbr()` returns a fixed public API and every test drives `start_service()`. A private helper would be no more testable than the inline block; making it genuinely testable would mean exporting internals, which is a substantially larger change than the one under review.

The block also isn't sitting in `create_pbr()` directly — it's inside `policy_process()`, immediately alongside the existing `errorPolicyNoSrcDest`, `errorPolicyNoInterface` and `errorPolicyUnknownInterface` guards. It follows the surrounding idiom exactly. Nine lines of guards among peer guards isn't what makes that file long.

Happy to extract it if you'd still prefer, but I'd rather not add indirection for a benefit the test setup can't realise.

### 2. Matching UI entries — confirmed present in luci-app-pbr#39

Both keys are in `status.js` (lines 302 and 305), and the sets match exactly:

```
pbr pushes:   ChainNft  DestPort  Proto  SrcPort
pkg.uc holds: ChainNft  DestPort  Proto  SrcPort
luci renders: ChainNft  DestPort  Proto  SrcPort
```

Worth flagging a false positive I hit while checking this: a naive `grep warningTor` over `pbr.uc` still matches `warningTorUnsetParams`, but only because the removed name appears in the **history comment** at `pbr.uc:648`. It is not defined in the catalog and never pushed.

On enforcement — that's what the ⚠️ merge-together note on both PRs is for. `isVersionMismatch()` catches it at runtime: merging either alone surfaces the version-mismatch warning until the other lands.

### 3. `running_iptables` — no consumers remain

Searched `pbr`, `luci-app-pbr` and `docs.mossdef.org`: **zero references**. The producer and both consumers go together —

- producer: `pbr.uc` status payload (this PR)
- consumers: the `"(iptables mode)"` branches in `status.js` and `72_pbr.js` (luci-app-pbr#39)

Both were unreachable anyway: the field has been hardcoded `false` since the iptables backend was removed in 1.1.7, so neither branch could ever be taken. Nothing needs redirecting to `running_nft` / `running_nft_file` — those branches already handle every reachable case.

### 4. Compat 36 — all five locations, no stragglers

| Location | Value |
|---|---|
| `files/lib/pbr/pkg.uc` | `compat: '36'` |
| `files/etc/init.d/pbr` | `initCompat='36'` |
| `tests/04_policies/01_start_dynamic_routing` | `packageCompat == 36` |
| luci `status.js` | `LuciCompat` → `36` |
| luci `luci.pbr` | `rpcdCompat = 36` |

Nothing left at 35. Note two of these are easy to miss with a plain grep — `initCompat` in the shell init script, and `LuciCompat`, which is a multi-line getter.

Nothing else in either repo reads compat: the only other matches are `errorIncompatibleUserFile` (unrelated — it's about user include files) and `rpcdCompat` being assigned into the RPC result. Neither `Makefile` references it, so packaging is unaffected.

No `PKG_RELEASE` bump in either PR — that's deliberate and will be handled separately.

egc112:
Co-P:ilot comment:

> Thanks — this looks good to me and the maintainer's reply confirms the important points (new warning keys present in luci-app-pbr#39, no remaining consumers of running_iptables, compat bumped everywhere, tests green).
> 
> Please merge these two PRs together to avoid a transient version-mismatch / “Unknown warning” experience for users:
> 
>     mossdef-org/pbr#150 (this PR)
>     mossdef-org/luci-app-pbr#39 (UI strings / compat)
> 
> After merging both, please coordinate the packaging step (PKG_RELEASE bump / release packaging) as a follow-up so the new compat value and strings are correctly shipped together.
> 
> I approve pbr#150 contingent on merging it simultaneously with luci-app-pbr#39 — please confirm when both are merged or if you want me to help with the packaging change.

PR #152, opened 2026-08-14
pbr-1.2.3: rename version check test.sh to test-version.sh

egc112:
Renames the root `test.sh` to `test-version.sh`. Pure rename — the script's contents are unchanged and stay byte-identical to the feed's copy.

## Why

openwrt/packages commit [`7cb0c183`](https://github.com/openwrt/packages/commit/7cb0c183a9e98f71366c462f9a38c13246003764) ("tree-wide: rename version check test.sh to test-version.sh") already renamed `net/pbr/test.sh` → `net/pbr/test-version.sh` across 17 packages on `master`. This brings our copy back in sync.

The two names are **not** interchangeable in `actions-shared-workflows`' `test_entrypoint.sh`:

- `test.sh` — package-specific test. Always run when present, and its presence *suppresses* the generic tests unless `force_generic_tests` is set.
- `test-version.sh` — version-check override, reached only from **inside** the generic tests. When present, per-binary version probing is skipped and this script runs instead.

Both are invoked as `sh <script> "$PKG_NAME" "$PKG_VERSION"`, which is why the one-liner greps `"$2"` — so the rename is behaviour-preserving for the check itself.

## Impact

- Nothing in this repo, luci-app-pbr, or the docs references `test.sh`; our own CI runs `tests/run_tests.sh`, which is unrelated and untouched.
- No `PKG_RELEASE` bump — the upstream rename touched no Makefiles.
- Release branches (`openwrt-25.12`, `openwrt-24.10`) still carry `test.sh`, but they pin `actions-shared-workflows@main`, so the new name is understood there too. Backporting 1.2.3 is safe whether the sync replaces the directory or only adds files.

Main benefit: with the old name gone, the outcome no longer depends on how the feed sync is performed — a wholesale-copy sync can no longer reintroduce `test.sh` into the feed, where it would suppress the generic tests if `force_generic_tests` were ever turned off.


egc112:
Co-pilot comment:

> LGTM — pure rename (test.sh → test-version.sh) to match upstream; script contents unchanged. Approving.

PR #154, opened 2026-08-14
pbr-1.2.2: bump PKG_RELEASE from 20 to 22

### LINKED ISSUE #130: [pbr] configuration breaks on sysupgrade
I have several custom user scripts uder the `/etc/pbr/` directory. After sysupgrade these scripts stop working because paths to them are changed to `/usr/share/pbr/...` in the pbr config.

**To reproduce**

1. Create custom user script and store it in `/etc/pbr/foo.user`.
2. `echo /etc/pbr/ >> /etc/sysupgrade.conf` to keep this script when upgrading.
3. Add this script to pbr configuration via `luci-app-pbr` interface and ensure it works.
4. Flash new sysupgrade image.

**Expected behavior**

Script works after reboot.

**Diagnostic info**

Sorry, but I can't paste full `service pbr support` output because it contains too much private information about VPN setup and routing.

```
# cat /etc/os-release 
NAME="OpenWrt"
VERSION="25.12.5"
ID="openwrt"
ID_LIKE="lede openwrt"
PRETTY_NAME="OpenWrt 25.12.5"
VERSION_ID="25.12.5"
HOME_URL="https://openwrt.org/"
BUG_URL="https://bugs.openwrt.org/"
SUPPORT_URL="https://forum.openwrt.org/"
FIRMWARE_URL="https://downloads.openwrt.org/"
BUILD_ID="r33051-f5dae5ece4"
OPENWRT_BOARD="ath79/generic"
OPENWRT_ARCH="mips_24kc"
OPENWRT_TAINTS=""
OPENWRT_DEVICE_MANUFACTURER="OpenWrt"
OPENWRT_DEVICE_MANUFACTURER_URL="https://openwrt.org/"
OPENWRT_DEVICE_PRODUCT="Generic"
OPENWRT_DEVICE_REVISION="v0"
OPENWRT_RELEASE="OpenWrt 25.12.5 r33051-f5dae5ece4"
OPENWRT_BUILD_DATE="1782737960"
```

`pbr` version is 1.2.2-r18.

The corresponding pbr configuration entry is

```
config include
        option enabled '1'
        option path '/etc/pbr/foo.user'
```

After sysupgrade it changed to

```
config include
        option enabled '1'
        option path '/usr/share/pbr/foo.user'
```

The error reported by the `pbr` initscript is `ERROR: Custom user file '' not found or empty!`.

**Suggested solution**

I see that the path is changed by [this line](https://github.com/mossdef-org/pbr/blob/84a29d4add9c1a479b6e3331fffa9182799a8df6/files/etc/uci-defaults/90-pbr#L21). If I understand correctly, this is done because user scripts packaged together with `pbr` were moved from `/etc/pbr` to `/usr/share/pbr`. However the correct place for custom configuration is still in `/etc`, according to FHS. User can not only place his own script here, but also copy a script from `/usr/share/pbr` and modify it without changing its' name. So my suggestion is to check file existance in `/etc/pbr` and edit path only if it is missing.

egc112:
### What's new in release 22

**Fixed: pbr could knock out LAN forwarding and port forwards.**
When rebuilding its rules, pbr flushed the firewall's own `forward`, `dstnat`, `prerouting` and `output` chains instead of its own — wiping fw4's zone jumps and port forwards. Normally a reload seconds later papered over it, but if that reload was skipped because the new ruleset failed validation, the outage stayed. Since `forward`'s default policy is drop, that meant no LAN↔WAN traffic at all. Backport of the fix already in 1.2.3.

**Fixed: custom script paths no longer rewritten on upgrade.**
Scripts kept in `/etc/pbr/` had their paths silently rewritten to `/usr/share/pbr/` on every upgrade, so after a sysupgrade they stopped being found. (#130)

**IPsec/xfrm interfaces are now recognised.** They count as tunnel interfaces like WireGuard and OpenVPN, and point-to-point detection is now done the same way everywhere — so no more spurious "unknown gateway" warnings on those interfaces.

**Fixed in the WebUI: your resolver set choice is no longer lost when you hit Save.**
Only affects LuCI master/snapshot builds at the moment — stable releases do not yet carry the luci-base change that triggers it, so on 24.10 and 25.12 you are unaffected. On master, the first Save on the overview page deleted the provisioned `resolver_set` value, silently turning resolver set support off while the WebUI still showed it enabled, and domain policies quietly fell back to one-off lookups. On 1.2.2 nothing restored it afterwards, so it stayed off. This fix lands ahead of that change reaching stable. Thanks to @pesa1234 for spotting it.

**Release/packaging.** apk files keep their `r` release prefix so downloads match locally built ones, stale release assets are pruned, the SDK is pinned to stable point releases, and new releases now sync to repo.mossdef.org immediately instead of waiting for the hourly poll.

The compatibility number is unchanged at 27 — nothing since release 20 touched the message catalog, so no forced pairing this time. Both packages are released together as usual.

---

**pbr commits since 20:** #131 `/etc/pbr/` path rewrite (fixes #130) · #140 xfrm/IPsec interface support · #145 fw4 chain flush fix · maintainer metadata
**luci-app-pbr commits since 20:** mossdef-org/luci-app-pbr#38 resolver_set preserved on save · release workflow and packaging fixes · maintainer metadata

Pairs with the luci-app-pbr 1.2.2 bump to 22; both should land together.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=pbr
 PKG_VERSION:=1.2.2
-PKG_RELEASE:=20
+PKG_RELEASE:=22
 PKG_LICENSE:=AGPL-3.0-or-later
 PKG_MAINTAINER:=Stan Grishin <stangri@melmac.ca>, Erik Conijn <egc112@msn.com>
 


egc112:
Co_pilot comment:

> Approved — LGTM.
> 
> This PR only bumps PKG_RELEASE from 20 → 22. The paired luci-app-pbr bump is already merged and CI is green, so I approve — please merge.
>

PR #156, opened 2026-08-17
pbr-1.2.3: record system() calls in mocklib, guard three regressions

egc112:
Test infrastructure only — nothing under `files/` is touched, and `tests/` is not shipped in the package, so this needs no `PKG_RELEASE` bump and no `luci-app-pbr` counterpart.

pbr issues much of its effect through `system()`: `nft_call()` runs `nft ...` via `sh.run()`, and the ip rule cleanup loops call `system()` directly. The mock in `tests/lib/mocklib.uc` traced those calls to the debug log and dropped them, so no testcase could ever see them — leaving chain flushes, rule deletions and teardown ordering untestable, and three shipped regressions without a guard.

## `tests: record system() calls in mocklib`

Records every command the mock is handed and exposes it next to the existing capture helpers:

```
mocklib.commands()             every command, in call order, as a copy
mocklib.commands('substring')  only the ones containing it
mocklib.commands(/regexp/)     only the ones matching it
mocklib.has_command(needle)    the same selection, as a boolean
mocklib.clear_commands()       drop the recording, to assert one phase
```

Array argv is joined with spaces before recording and nothing is filtered out, including the `[ -t 2 ]` tty probe, so the list reflects the process as it ran. `sh.run()` appends ` >/dev/null 2>&1` to everything, so testcases match on substrings or anchored regexps rather than equality. Additive: no existing mocklib behavior changes.

The accompanying testcase covers the API itself, then asserts over a real `start_service()` run — which doubles as end-to-end proof that the recorder sees calls leaving no other trace.

## Three guards it unlocks

**`tests: guard nft cleanup against flushing fw4's base chains`** — regression guard for #143 (`bd042da`). `cleanup()`'s `'main_chains'` case flushed bare `forward` / `output` / `prerouting` / `dstnat` instead of the `pbr_`-prefixed chains, so every start flushed fw4's own base chains, taking down LAN↔WAN forwarding and every NAT port forward until fw4 was reloaded. The fix went in unguarded — nothing in the suite matched flush against chain names. Asserts a start flushes exactly the four `pbr_`-prefixed chains and never a bare one, never flushes the table, and flushes rather than deletes, since fw4 owns their definitions.

**`tests: guard the ip rule cleanup window against priority 0`** — regression guard for `12b9937`. `cleanup()`'s `'main_table'` window runs from `uplink_ip_rules_priority + 1` down to `uplink_ip_rules_priority - (fw_mask / uplink_mark)`. With a widened mask that lower bound goes negative and the window swallows priority 0 — the kernel's own `from all lookup local` rule. Deleting it kills local delivery: the router stops answering on its own addresses and needs a physical reset. Uses the config from the fix's own comment (`uplink_ip_rules_priority` 99, `fw_mask` ffff0000, unclamped bound −65436).

**`tests: assert mwan4 survives a pbr stop`** — the README promises pbr and mwan4 can share a router, but `04_policies/07_policy_mwan4_strategy` covers strategy targeting only and never calls `stop_service`, so nothing asserted that teardown leaves mwan4 alone. Runs a full start then a stop with mwan4 installed, and asserts the stop touches neither its chains nor its ip rules, that no chain operation reaches a chain pbr does not own, and that fw4's base chains and the table survive. Phrased around what pbr *may* touch rather than today's exact teardown, so a future change that does clean pbr's own chains on stop still passes.

## Verification

Each guard was checked against a simulated regression, not just against passing code:

- base-chains guard: reintroducing the unprefixed flush fails the relevant checks
- priority-0 guard: dropping the clamp floor fails 3 of 11 checks, including both priority-0 deletes
- mwan4 guard: sweeping mark chains without the `pbr_mark_` prefix filter and adding `'marking_chains'` to the stop cleanup fails 3 of 11 checks

Suite is 51/51 on this branch.

### DIFF
--- a/tests/01_validation/04_mocklib_system_recorder
+++ b/tests/01_validation/04_mocklib_system_recorder
@@ -1,0 +1,83 @@
+Testing the mocklib system() recorder itself. mocklib.commands() must return
+every command system() was called with, in call order, optionally filtered by
+substring or regexp, and as a copy the caller cannot disturb; has_command()
+mirrors it as a boolean; clear_commands() resets the recording so a test can
+assert on one phase at a time.
+
+The second half runs a real start_service() to show the recorder sees the
+calls that leave no other trace: the nft chain flushes issued through
+sh.run() and the 'ip rule del' loops that call system() directly. Neither
+appears in /var/run/pbr.nft nor in the start_service() result, so before this
+recorder existed there was no way to assert on them at all.
+
+-- Testcase --
+import create_pbr from 'pbr';
+
+let mocklib = global.mocklib;
+
+// ── the recorder API on its own ─────────────────────────────────────
+mocklib.clear_commands();
+let empty = mocklib.commands();
+
+system('/usr/libexec/ip-full -4 rule del priority 0 2>/dev/null');
+system('nft flush chain inet fw4 pbr_forward >/dev/null 2>&1');
+system(['/usr/libexec/ip-full', '-6', 'rule', 'del', 'priority', '0']);
+
+let all = mocklib.commands();
+let substring_hits = mocklib.commands('rule del priority 0');
+let regexp_hits = mocklib.commands(/^nft /);
+
+// mutating what commands() handed back must not disturb the recording
+let copy = mocklib.commands();
+push(copy, 'never executed');
+
+let checks = [
+	['clear_empties',        length(empty) == 0],
+	['records_every_call',   length(all) == 3],
+	['records_in_order',     index(all[0], '-4 rule del') >= 0 &&
+	                         index(all[1], 'nft flush') >= 0],
+	['array_argv_joined',    all[2] == '/usr/libexec/ip-full -6 rule del priority 0'],
+	['substring_filter',     length(substring_hits) == 2],
+	['regexp_filter',        length(regexp_hits) == 1 &&
+	                         index(regexp_hits[0], 'pbr_forward') >= 0],
+	['returns_a_copy',       length(mocklib.commands()) == 3],
+	['has_command_true',     mocklib.has_command('nft flush chain')],
+	['has_command_false',    !mocklib.has_command('fw4 -q reload')],
+	['has_command_regexp',   mocklib.has_command(/rule del priority 0$/)],
+];
+
+// ── the same recorder over a real start ─────────────────────────────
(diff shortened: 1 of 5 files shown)


egc112:
Thanks — looked at both, and neither needs a code change. Details below, plus one comment added in 436a517.

**1. `return 0` for non-tty commands**

Already there. The function ends:

```ucode
	/* Return 1 for tty checks so output goes to logger, not stderr */
	if (type(argv) == "string" && index(argv, "[ -t ") >= 0)
		return 1;

	return 0;
};
```

The diff hunk ends just before it, so it isn't visible in the patch context. It is also pre-existing — the only line this PR adds inside `global.system` is the `push(_commands, ...)` call, which doesn't alter the return path.

**2. Preserving and restoring the original `global.system`**

Declining this one, for two independent reasons.

*It isn't needed.* `run_testcase()` spawns a fresh `ucode` process per testcase (`tests/run_tests.sh:172`) with `-l mocklib` loaded into that process. Every global — the override and `_commands` alike — lives and dies with it, so there is no cross-suite contamination to restore from. That is also why `clear_commands()` exists: for asserting a single *phase* within one run (clear after `start_service()`, then assert on `stop_service()`), not for cleaning up after another test.

*It would weaken the harness.* There is no "previous mock behaviour" to restore — `global.system` before the override is the ucode builtin, which really executes commands. A `restore_system()` helper would give a test the ability to run `nft delete set`, `ip rule del` and `fw4 -q reload` against the developer's own machine, which is precisely what the mock exists to prevent. Nothing in the suite asserts on real execution; every test reads recorded commands or captured file contents.

I have added a comment on `_commands` in 436a517 making the per-process lifetime explicit, so the question doesn't come up again:

```ucode
/* Commands passed to system() — used by tests to assert on what was executed.
   run_tests.sh spawns one ucode process per testcase, so this starts empty for
   each and cannot leak into the next; clear_commands() is for asserting a single
   phase within one run, not for cleaning up after another test. */
```

Suite is 51/51 with that change.

egc112:
Co-pilot review:
`Short version — looks good to merge once CI runs the test suite: the author addressed the two review points (the system() return path already exists, and they explained why restoring the original global.system is unnecessary and added a comment). No remaining blockers from my side.`

PR #158, opened 2026-08-18
pbr-1.2.3: empty a resolver set when its policy's domain list changes

egc112:
## The bug

`nft` set names are keyed on the policy's uid, not on its contents, so editing a policy's `dest_addr` reuses the same set. Since #155 stopped deleting live sets on reload, that set survives the rebuild still holding every address the **old** domain list resolved to — and the policy goes on routing a domain that is no longer configured anywhere. Not in `/etc/config/pbr`, not in `/var/run/pbr.dnsmasq`, not in the ruleset. Only in the set the rule matches on, and nothing logs it.

`cleanup('orphan_sets')` does not catch this: the set is still declared by the new ruleset, so it is kept. That is why *disabling* a policy did clear its set — a disabled policy declares nothing, so the set is reaped as an orphan — while *editing* one did not. The two operations appeared to behave inconsistently for no visible reason, which is how this was found.

`nft_set_timeout` / `nft_set_gc_interval` are unset by default, so the entries are permanent until reboot.

Nothing in the package ever emptied a live set. `nftset.flush()` existed for precisely this and had no callers.

## The fix

`resolver.store_hash()` now also snapshots which domains fed which set, parsed out of the dnsmasq file before `configure()` truncates it — it already read that file for the md5, so the previous generation was available and simply not kept. After `nft_file.apply('main')`, the new `resolver.flush_changed()` re-parses the regenerated file and empties only the sets whose domain list actually differs.

Deliberately narrow, on both sides:

- **Flush, never delete.** The set object never goes away, so dnsmasq's `nftset=` target stays resolvable and no reply lands in a gap. That gap is what #155 closed and it is not being reopened. Flushing a set referenced by live rules is always permitted; only deleting one is refused with `EBUSY`.
- **A live `nft` command after `apply('main')`, not a line in `30-pbr.nft`.** `fw4` re-executes that file on every reload, including reloads `pbr` did not trigger, so a `flush set` baked into it would empty the sets continuously.
- **Only sets present in both generations with a differing domain list.** Flushing everything on every reload would undo #155 from the other side: a domain a client still holds cached is never re-queried, so it would silently drop out of policy routing. A set with no previous generation to compare against is left alone for the same reason.
- **Before `resolver.restart()`**, so addresses re-resolved after the restart are not wiped by a flush that follows.

The comparison is a sorted, de-duplicated `domain,domain` string per set name, so a rewrite that merges or orders the file differently without changing which domains feed a set still compares equal.

`nftset.flush()` now delegates to a new `nftset.flush_name()`, which flushes by set name, rather than repeating the command.

## Reload paths

`service pbr reload` arrives as `on_reload` and resolves to the same `default:` branch as `on_start`, so it does run this — and is proven inert when nothing changed.

`on_interface_reload` takes its own branch, which reloads routing only: it never calls `configure()` or `nft_file.apply('main')`, so it cannot reach the flush at all.

## Tests

`tests/03_nft_rules/09_flush_changed_domain_sets` seeds a previous generation and covers three policies: a changed domain list flushes both families, an identical one is left alone, a set with no previous generation is untouched. It also asserts the set is flushed rather than deleted. Verified as a real guard — with the `flush_changed()` call removed it fails on the two changed-set assertions instead of passing vacuously.

`tests/03_nft_rules/10_reload_keeps_unchanged_sets` pins the two paths that must never empty anything.

Its interface-reload half needed a new fixture. `get_mark_nft_chains()` runs `nft list table inet fw4 2>/dev/null`, but the only popen fixture in the tree was for `2>&1`, so `is_service_running_nft()` was always false in tests and any `on_interface_reload` case was silently downgraded to `on_start` — testing the wrong branch while still passing. The test now also asserts `nft_file.apply('main')` never ran on that path, so it cannot go vacuous again.

54/54 passing.

## Verification

Tested on the router.

### DIFF
--- a/files/lib/pbr/nft.uc
+++ b/files/lib/pbr/nft.uc
@@ -21,6 +21,11 @@ function create_nft(fs_mod, config, sh, output, pkg, platform, network, V, state
 	let nft_fw4_dump = '';
 	let resolver_working_flag = false;
 	let resolver_stored_hash = '';
+	// Per-set domain lists as of the previous generation of the dnsmasq file,
+	// snapshotted by store_hash() before configure() truncates it. Null means
+	// "no previous generation seen", which is deliberately not the same as
+	// "empty" -- see resolver.flush_changed().
+	let resolver_prev_domains = null;
 	let dnsmasq_ubus = null;
 
 	// Forward declaration (circular: resolveip_to_nftset → resolver → nftset → resolveip_to_nftset)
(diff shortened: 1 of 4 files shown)


egc112:
Co-pilot comment:

> 
> Status: Low risk, ready to merge
> 
>     Clear intent: Bug is well-documented; fix is surgical and narrow
>     Test coverage: 8/8 assertions in 09_flush_changed_domain_sets pass; new test guards against regression if flush_changed() is removed
>     No side effects: Flush only happens when domain lists actually differ; unchanged sets keep their addresses (preserving #155's benefit)
>     Reload paths protected: Both on_reload and on_interface_reload have explicit guards tested in 10_reload_keeps_unchanged_sets
>     Code quality: Well-commented, consistent with codebase style; state management is explicit (null vs. empty distinction)
> 
> Observations:
> 
>     No blocking issues
>     Clean merge (1 commit, 4 files changed: 313 additions, 4 deletions)
>     CI/checks status: No reviews requested yet
>     Author tested on real hardware
>

PR #159, opened 2026-08-18
pbr-1.2.3: apply address negations on the policy's own rules

egc112:
## The bug

A negated entry in `src_addr`/`dest_addr` got a rule of its own, one per `*_negative` `filter_options` group:

```
ip saddr != { 10.0.0.22 } counter packets 24 bytes 1991 goto pbr_mark_0x010000 comment "phone-to-wan"
```

`nft` has no OR between the match expressions of a rule, so `pbr` needs one rule per positive address type — that is what makes a policy's address list a **union** of sources. Negated entries are the opposite: they are **AND** terms. On a rule of its own, `!10.0.0.22` reads as "every source except 10.0.0.22" — a catch-all that matches nearly everything, routes the whole network, and swallows exactly the traffic the policy was meant to leave alone. The counter above is from a real router: that rule took the entire chain's traffic, from an interface the policy does not even name.

`src_port`/`dest_port` negation was already built inline on the rule it belongs to, which is why `dest_port='!80'` always behaved correctly.

## The fix

Each policy's negated groups are resolved once, in the new `collect_negations()`, into the fragments appended to every rule the policy emits. `policy_process()` and `dns_policy_process()` now iterate the positive groups only and hand those fragments down to `policy_routing()` / `dns_policy_routing()`.

`classify_addr()` already keeps the families apart, so an IPv4 exclusion only ever reaches the IPv4 rule and vice versa. `matched` feeds the unknown-entry validation, so negated tokens are still not reported as unknown.

For `src_addr '192.168.9.0/24 98:B8:BC:8B:3F:9E @br-lan !11:22:33:44:55:66 10.0.0.0/24 !10.0.0.22'`, from the router:

```
iifname { "br-lan", "br-guest" } ether saddr != 11:22:33:44:55:66 ip saddr != 10.0.0.22 goto pbr_mark_0x010000
meta nfproto ipv6 iifname { "br-lan", "br-guest" } ether saddr != 11:22:33:44:55:66 goto pbr_mark_0x010000
ether saddr { 00:11:22:33:44:55, 98:b8:bc:8b:3f:9e } ether saddr != 11:22:33:44:55:66 ip saddr != 10.0.0.22 goto pbr_mark_0x010000
meta nfproto ipv6 ether saddr { 00:11:22:33:44:55, 98:b8:bc:8b:3f:9e } ether saddr != 11:22:33:44:55:66 goto pbr_mark_0x010000
ip saddr { 10.0.0.0/24, 192.168.9.0/24 } ether saddr != 11:22:33:44:55:66 ip saddr != 10.0.0.22 goto pbr_mark_0x010000
```

Positives OR across rules, exclusions AND within each.

### Why the family pinning

An `iifname` or `ether saddr` match reads the same for both families, so it is normally emitted once (the `param4 == param6` dedup). Attaching an IPv4-only exclusion splits it, and the two copies then **have** to be pinned with `meta nfproto` — otherwise the IPv6 copy, carrying no IPv4 exclusion, re-admits the very packets the IPv4 copy just filtered out, and the exclusion silently does nothing.

The guard is added only when the positive part is non-empty. A rule built from exclusions alone instead suppresses the family that has none, which keeps the documented standalone form (`dest_port '!80'`, "everything except these") behaving as before. `nft` does not print the redundant `meta nfproto ipv4` where an `ip` match already implies it, which is why only the IPv6 copies show it above.

## Second bug, fixed here too

For a negated domain, `classify_addr()` referenced a set named `<set>_neg` while `create_set()` and the `nftset=` line were keyed on the un-suffixed uid — so the set the rule pointed at was never created and never populated. That has been true since the shell implementation, where `nftset()` built the name from `$uid` and only the caller appended `$nftset_suffix`.

A policy mixing positive and negated domains was worse than ineffective: **all** of its domains, the excluded ones included, landed in the positive set, and the dangling `@<set>_neg` reference failed the `nft -c -f` check in `nft_file.apply()` — so the entire `pbr` ruleset was never installed, taking every other policy down with it.

The suffix is folded into the uid and the reference built with `get_set_name()`, so creation, dnsmasq population and the rule all derive one name. Confirmed on the router:

```
add set inet fw4 pbr_wan_4_dst_ip_cfg1a6ff5_neg { type ipv4_addr; ... }
nftset=/downloads.openwrt.org/4#inet#fw4#pbr_wan_4_dst_ip_cfg1a6ff5_neg,...
ip daddr @pbr_wan_4_dst_ip_cfg1a6ff5 ip daddr != @pbr_wan_4_dst_ip_cfg1a6ff5_neg goto pbr_mark_0x010000
```

The suffix is needed rather than removable: set names are keyed on the policy uid, not on the group within it, so without it a policy holding both a positive and a negated domain would emit `ip daddr @S ip daddr != @S`, which can never match.

## Upgrade path

On the old code a negated domain was fed into the **positive** set, so this fix changes that set's domain list — and #158's flush therefore empties it on the first reload after upgrade, clearing the wrongly-routed addresses by itself. No manual intervention, no reboot. That is why #158 went first.

## Tests

`tests/03_nft_rules/11_negation_on_same_rule` — 19 assertions: rule counts, every rule carrying a positive match, per-family exclusion placement, the `meta nfproto` pinning (present on the split copies, absent where an `ip` match already implies it), `dest_addr` exclusions, the negated-domain set being created *and* populated *and* referenced consistently, and a guard that no bare exclusion rule exists anywhere except the documented standalone form.

`tests/lib/mocklib/fs.uc` had no `lstat()` at all, so `is_phys_dev()` always returned false and `@device` entries could not be tested. `stat()` is hoisted to a module-level `stat_path()` with both aliased to it — it must not be a method, since `pbr` passes `fs.lstat` around as a bare function reference and would lose the receiver.

55/55 passing.

## Verification

Tested on the router: rule shapes, counters, the `_neg` sets being created and populated by dnsmasq, and the negated-domain policy loading where it previously failed the ruleset check.

### DIFF
--- a/files/lib/pbr/nft.uc
+++ b/files/lib/pbr/nft.uc
@@ -1168,6 +1168,10 @@ function create_nft(fs_mod, config, sh, output, pkg, platform, network, V, state
 		// different list of domains than the policy's positive group, and both
 		// are keyed on the same uid. Fold the suffix into the uid so the set
 		// that gets created and populated is the very one the rule references.
+		// A falsy uid is passed through untouched: only the resolver branch
+		// below consumes this, and the one caller that passes no uid -- the DNS
+		// path, with use_resolver false -- never reaches it, so there is no set
+		// name to suffix.
 		let set_uid = (uid && nftset_suffix) ? '' + uid + nftset_suffix : uid;
 		let first_val = V.str_first_word(value);
 		let param4 = '', param6 = '';
--- a/files/lib/pbr/nft.uc
+++ b/files/lib/pbr/nft.uc
@@ -1172,7 +1172,8 @@ function create_nft(fs_mod, config, sh, output, pkg, platform, network, V, state
 		// below consumes this, and the one caller that passes no uid -- the DNS
 		// path, with use_resolver false -- never reaches it, so there is no set
 		// name to suffix.
-		let set_uid = (uid && nftset_suffix) ? '' + uid + nftset_suffix : uid;
+		let set_uid = uid;
+		if (uid && nftset_suffix) set_uid = '' + uid + nftset_suffix;
 		let first_val = V.str_first_word(value);
 		let param4 = '', param6 = '';
 		let empty4 = false, empty6 = false;
--- a/files/lib/pbr/nft.uc
+++ b/files/lib/pbr/nft.uc
@@ -1164,6 +1164,11 @@ function create_nft(fs_mod, config, sh, output, pkg, platform, network, V, state
 			value = replace(value, /!/g, '');
 			nftset_suffix = '_neg';
 		}
+		// The negated group of a policy needs its own resolver set: it holds a
+		// different list of domains than the policy's positive group, and both
+		// are keyed on the same uid. Fold the suffix into the uid so the set
+		// that gets created and populated is the very one the rule references.
+		let set_uid = (uid && nftset_suffix) ? '' + uid + nftset_suffix : uid;
 		let first_val = V.str_first_word(value);
 		let param4 = '', param6 = '';
 		let empty4 = false, empty6 = false;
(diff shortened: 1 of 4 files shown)


egc112:
Thanks for the review — the `s.fg` / `d.fg` finding is a false positive, and the suggested fix would break the build. Verified rather than argued:

**`for...in` over an array yields values in ucode, not indices.** This is a deliberate divergence from JavaScript:

```
$ ucode -e 'let a=[{fg:"ipv4",fv:"x"},{fg:"ipv6",fv:"y"}]; for (let s in a) printf("type=%s fg=%s fv=%s\n", type(s), s.fg, s.fv);'
type=object fg=ipv4 fv=x
type=object fg=ipv6 fv=y
```

`s` is the object; `s.fg` and `s.fv` are exactly what was pushed.

**ucode has no `for...of`:**

```
$ ucode -e 'let a=[1,2]; for (let s of a) print(s);'
Syntax error: Unexpected token
Expecting ';'
```

Applying the suggested one-line change would make the module fail to parse, so no policy rules would be generated at all — the outcome the comment warns about.

**Corroboration.** It is the existing idiom throughout, all of this pre-dating this PR:

```
pbr.uc:313   for (let proto_i in protos) {          // proto_i is 'tcp' / 'udp'
pbr.uc:499   for (let dest_rule in tor_rules) {     // concatenated as a string
nft.uc:53    for (let line in split(out, '\n')) {   // match(line, ...) on the value
```

`grep` finds no `for...of` anywhere under `files/lib/pbr/`.

It is also load-bearing in the tests: were `s` an index, `policy_routing(..., s.fv, ...)` would pass `undefined` as `src_addr` and emit no rules, so `11_negation_on_same_rule` would fail on all five of the generated rules it asserts. CI is green at 55/55, and the five rules were confirmed on real hardware.

That said, the reading is an easy one to make, so I have designed it out in a3b4e3a: the variables are now `src_group` / `dest_group`, with a comment noting the iteration semantics. No behaviour change.

egc112:
Co-pilot comment:

> LGTM — the update addresses the iteration/readability concerns and the set-name handling; I found no blocking issues.

PR #161, opened 2026-08-18
pbr-1.2.3: keep the second uplink's chain name in step with its mark

egc112:
## The bug

With `uplink_interface4 != uplink_interface6`, the second uplink reached shares the first's firewall mark — they share a routing table, so they must — but its chain name was derived from the mark it held *before* that reassignment and never recomputed. The netifd and mwan4 branches beside it both recompute; only this one does not. `split_uplink_second` suppresses the mark and priority increment and nothing else.

So the second uplink names the chain of whichever interface really owns the mark it no longer uses. From a live router:

```
define pbr_wan_chain         = pbr_mark_0x010000
define pbr_wan6_chain        = pbr_mark_0x020000   <- stale
define pbr_wg_mullv_se_chain = pbr_mark_0x020000   <- real owner
```

## Two ways it goes wrong, both reproduced

**Another interface owns the mark.** A policy, an `<iface>_dscp` option or `icmp_interface` pointed at the second uplink emits a `goto` at that interface's chain, so the traffic takes its mark and leaves by its table. Confirmed on a DL-WRX36: with `wan6_dscp=10` and a WireGuard tunnel present, DSCP-10 traffic was marked `0x020000` and routed out the tunnel instead of wan6. The ruleset installs cleanly and nothing is logged.

**Nothing owns it** — a plain `wan` + `wan6` router with no tunnels. `ensure_mark_chain()` dedupes on the mark, so the second uplink returns early and no chain is created under its stale name. The ruleset then carries a `goto` at a chain that does not exist, fails the `nft -c -f` check in `nft_file.apply()`, and **pbr installs no rules at all**. Two config options reach this and no policy is involved.

## Not specific to the IPv6 uplink

The branch keys on order of enumeration, so which uplink is "second" follows section order in `/etc/config/network`. With `wan6` listed first the stale name lands on **wan**, where an ordinary `option interface 'wan'` policy leaves by a tunnel.

## Scope is exactly this one field

Priorities and table ids are unaffected — both uplinks copy the same `_uplink_priority` and the counter only decrements when `!split_uplink_second`, and `interface_resolve_tid()` reads the other uplink's tid and reuses it. Diffing the emitted `ip rule` commands between the two enumeration orders gives identical sets with identical mark-to-priority pairings; only the sequence differs.

The non-split path already pairs every mark with its own chain, and netifd and mwan4 interfaces never reach this branch. `nft` accepts a `goto` at a chain declared later in the same batch (verified on the router), so file ordering alone is not a failure mode — the chain has to be genuinely absent.

## The fix

What the two branches above already do: recompute `_chain_name` when `_mark` changes.

## Tests

`tests/05_interfaces/13_split_uplink_chain_name` covers the first form — the uplinks share a chain, it is not the tunnel's, no two interfaces collide unexpectedly, no `goto` dangles, and the DSCP and ICMP rules target the uplink's chain. Without the fix it fails four of nine.

`tests/05_interfaces/14_split_uplink_no_dangling_chain` covers the second: every tunnel ignored, so nothing owns the mark. Exactly one `pbr_mark_` chain must exist and every `goto` target must be created. Without the fix it fails two of four.

Both drive it through `wan6_dscp` rather than a policy, to hold the tests to the fact that no policy is needed to reach this.

57/57 passing.

## Verification

Confirmed on the router: before the patch `nft list chain inet fw4 pbr_prerouting | grep dscp` gave `goto pbr_mark_0x020000` (the tunnel's chain); after, `goto pbr_mark_0x010000`, matching `define pbr_wan_chain`.

## Deliberately not included

Whether the IPv6 uplink should be selectable at all when `is_split_uplink()`. That is a design question, and reversed enumeration order weakens it as a remedy — the stale name can land on `wan`, which cannot be removed from the list.

### DIFF
--- a/files/lib/pbr/pbr.uc
+++ b/files/lib/pbr/pbr.uc
@@ -1113,6 +1113,15 @@ function create_pbr(fs_mod, uci_mod, ubus_mod) {
 				if (net.is_uplink4(iface) || net.is_uplink6(iface)) {
 					if (_uplink_mark && _uplink_priority) {
 						_mark = _uplink_mark;
+						// The chain name is derived from the mark above, so it
+						// has to follow the mark here as it does in the netifd
+						// and mwan4 branches. Left stale, the second uplink
+						// names the chain of whichever interface really owns
+						// the mark it no longer uses -- silently routing its
+						// traffic out that interface -- or, when no interface
+						// owns it, names a chain nobody creates, which fails
+						// the ruleset check and installs nothing at all.
+						_chain_name = pkg.nft_prefix + '_mark_' + _mark;
 						_priority = _uplink_priority;
 						split_uplink_second = true;
 					} else {
--- a/files/lib/pbr/pbr.uc
+++ b/files/lib/pbr/pbr.uc
@@ -1069,6 +1069,13 @@ function create_pbr(fs_mod, uci_mod, ubus_mod) {
 	// step: an interface whose mark is reassigned after the name was built ends up
 	// naming another interface's chain, or one nobody creates. Deriving both from
 	// here makes that structural rather than something each branch remembers.
+	//
+	// 'mark' is a '0x%06x' string. Every caller is already guaranteed one: it is
+	// either iface_mark, built by sprintf() below, or a netifd/mwan4 mark reached
+	// only through a branch that has tested it for truth. No coercion or fallback
+	// here on purpose -- a sentinel name would be created by nobody and routed to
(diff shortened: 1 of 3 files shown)


egc112:
Co-pilot comment:

> Conclusion / recommendation
> 
>     The main suggestion (centralize chain-name construction) has been applied — thank you, that addresses the largest stylistic/maintenance point.
>     Functionally this remains a small, safe fix and the tests cover the documented failure modes. This is low risk and ready to merge once CI runs/pass.

PR #164, opened 2026-08-22
pbr.user.dnsprefetch: make it work on the ucode service again

egc112:
`pbr.user.dnsprefetch` has been a silent no-op since the ucode rewrite, and reported success while doing nothing. Four commits: the fix, its header documentation, a warning for the one remaining silent path, and timeout budgets sized for slow hardware.

### Why it did nothing

pbr sources `.sh` user files from a bare shell that exports none of the service's internals (`pbr.uc` → `sys.run` → `system()`), so `$resolverSetSupported` was empty and `main()` bailed at its first gate. Because the body was backgrounded, that failure never reached `sh.run` either: pbr logged `Running ... [OK]` while nothing had happened, and the abort line went to syslog under an empty tag. It ships `enabled '0'`, which is why nobody reported it.

The trigger was broken independently. It armed on a dnsmasq restart via ubus, but since #155 a no-change reload takes the quieter SIGHUP cache flush, so the common path always timed out. Worse, pbr waits for every process a user file leaves running, so waiting for pbr's own progress deadlocks it — a 60s timeout made `/etc/init.d/pbr reload` take 62s instead of 3s.

### What it does now

Re-exec under `setsid` so the reload returns immediately, then watch the mtime of the installed ruleset, settle, and resolve. SIGHUP dnsmasq first so a cached name still reaches its set — but only once it answers a query, because the default action for SIGHUP is terminate and signalling it mid-restart killed the daemon outright on test hardware.

The script defines what it needs locally rather than inheriting it, and reads `pbr.config.resolver_set` for the gate `$resolverSetSupported` used to provide.

### Validation

On a DL-WRX36 running 1.2.3-r91, sets emptied before each run:

- no-change reload (SIGHUP branch) and forced restart — 6/6 domains resolved, sets refilled, reload unaffected
- self-path pointing at a missing file — warning logged, reload not slowed, service stays active
- two reloads 3s apart — both workers complete, no fifo or process leaks
- interface-reload trigger — does **not** prefetch, which is now documented; it does when pbr promotes the event to a full start, verified by clearing pbr's chains
- three reboots — nothing on the `on_boot` pass, nothing on a start deferred with the uplink down, 6/6 on each real start

Those boots also showed dnsmasq answering up to 33s after pbr logs `Restarting dnsmasq [OK]`, against a 30-iteration budget — hence the last commit.

Scoped to `files/usr/share/pbr/pbr.user.dnsprefetch`. Requires the ucode implementation; 1.2.2 keeps its own copy. No `PKG_RELEASE` bump here.

### DIFF
--- a/files/usr/share/pbr/pbr.user.dnsprefetch
+++ b/files/usr/share/pbr/pbr.user.dnsprefetch
@@ -6,6 +6,25 @@
 # on the shell implementation (1.2.2 and earlier), which ships its own copy of
 # this file -- use that one there.
 #
+# WHEN IT RUNS: once per full pbr run -- start, restart and reload. Not on the
+# on_boot pass, not on a start deferred because the uplink is down, and not on
+# stop. Overlapping runs are harmless: each detached copy makes its own
+# $$-suffixed fifo.
+#
+# An interface event (a wan flap) does NOT run it. That path reloads the
+# interfaces and returns; user files, the ruleset rewrite and the resolver step
+# all sit in the branch it skips. It only prefetches when pbr promotes the event
+# to a full start, which it does when its ruleset is not installed, when procd
+# reports no gateways, or when the last run recorded errors. So the sets are
+# refilled after the reload that follows a flap, not by the flap itself.
+#
+# There is no timer. Nothing re-resolves a name once its TTL expires and
+# dnsmasq drops it from the cache, so a set keeps the addresses it got until
+# the next pbr run. That is inherent to a prefetch rather than a refresher; a
+# domain that rotates addresses faster than pbr reloads needs a cron job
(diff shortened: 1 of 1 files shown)


egc112:
Co-pilot review:

> Status: ✅ Ready to merge — all edge cases validated on real hardware
> 
> The concerns raised in the initial review have been addressed with measured testing and solid reasoning:
> 
>     ✅ Dnsmasq readiness probe is now robust across nohosts configurations
>     ✅ Process cleanup is validated by measurement, not guesswork
>     ✅ Subtle timing issues (SIGHUP mid-restart) are guarded against
>     ✅ No silent failures remain
> 
> No blocking issues. The PR demonstrates the kind of careful, hardware-validated engineering that pays dividends in production. Merge when ready

PR #165, opened 2026-08-25
pbr-1.2.2: add symmetric return routing

tracyhatemice:
Replies follow the main routing table, so a connection arriving on a non-default uplink is answered out of whichever interface owns the default route. Inbound connections on a secondary WAN or a tunnel break as a result: port forwards never complete and services on the router are unreachable from that path.

Add an opt-in `symmetric_return` option that pins a connection's replies to the interface its first packet arrived on.

- New `symmetric_return` (bool, default off) and `symmetric_return_interface` (list) options.
- New `pbr_symmetric_prerouting` and `pbr_symmetric_output` chains, jumped from fw4's mangle_prerouting and mangle_output.
- New connections save their interface mark into `ct mark`; replies restore it into `meta mark`, selecting that interface's routing table.
- cleanup flushes the symmetric chains.
- status and support gain a `chains - symmetric return` section.

Nothing is emitted unless the feature is enabled, so a default install produces an unchanged ruleset.

Strict reverse path filtering drops the inbound packet before these rules are reached, so net.ipv4.conf.*.rp_filter must not be set to 1 on the interfaces involved. OpenWrt's default of 2 (loose) works.

Credit is extended to a contributor who wishes to remain anonymous.


egc112:
Thanks for your contribution unfortunately 1.2.2 is EOL so we are not adding any new code to this branch.

So while appreciating your contribution I am closing this PR

PR #166, opened 2026-08-26
pbr-1.2.3: add symmetric return routing

tracyhatemice:
This is a port of https://github.com/mossdef-org/pbr/pull/165 onto 1.2.3

Replies follow the main routing table, so a connection arriving on a non-default uplink is answered out of whichever interface owns the default route. Inbound connections on a secondary WAN or a tunnel break as a result: port forwards never complete and services on the router are unreachable from that path.

Add an opt-in `symmetric_return` option that pins a connection's replies to the interface its first packet arrived on.

- New `symmetric_return` (bool, default off) and `symmetric_return_interface` (list) options.
- New `pbr_symmetric_prerouting` and `pbr_symmetric_output` chains, jumped from fw4's mangle_prerouting and mangle_output.
- New connections save their interface mark into `ct mark`; replies restore it into `meta mark`, selecting that interface's routing table.
- cleanup flushes the symmetric chains.
- status and support gain a `chains - symmetric return` section.

Nothing is emitted unless the feature is enabled, so a default install produces an unchanged ruleset.

Strict reverse path filtering drops the inbound packet before these rules are reached, so net.ipv4.conf.*.rp_filter must not be set to 1 on the interfaces involved. OpenWrt's default of 2 (loose) works.

Credit is extended to a contributor who wishes to remain anonymous.


egc112:
Thanks for you contribution, at this moment we are only fixing bugs and not implementing new code as the new ucode port 1.2.3. should go public the coming weeks.

Furthermore the main developer @stangri is consumed by his day time job and he is the principal architect and knows more about mwan4. 
So for now I am parking this but leave it open for further review and possible implementation  in the future as I understand the problem you ran into.

Again thanks for contributing

stangri:
@tracyhatemice thank you for this invaluable contribution.

I wonder if you could address a few use case questions I have:
1. Would it make sense to have just one `symmetric_response` option, which when enabled, would enable it on all supported interfaces (and not require an additional list of interfaces to enable this on)? Can you estimate if there will be a performance penalty in this scenario?
2. This seems to obsolete the "default ICMP interface" feature, should we remove that?

egc112:
Just my two cents:
1. I think the cost is a `ct mark` write per new connection, a match on every reply-direction packet in both prerouting and output, and the reroute that fw4's `mangle_output` (`type route`) forces whenever the mark changes, so I would not make it global

2. Default ICMP interface has a different use case
`icmp_interface` emits into `pbr_output` only.
The commonest reason to set `icmp_interface` is unaffected: router has a VPN as default route, but its own pings must go out the real WAN so monitoring tests the real uplink. Those users frequently have no inbound connections at all — there is nothing for symmetric return to be symmetric about.

In general this is a must have for a multi-wan package, but PBR was originally not designed as such.
If you want to extend PBR's functionality to be a multiwan package then this is a very useful addition. So I think this functionality should be added to the mwan4 integration and should be operational if that is activated but I have not kept track of mwan4.

tracyhatemice:
> 1. Would it make sense to have just one `symmetric_response` option, which when enabled, would enable it on all supported interfaces (and not require an additional list of interfaces to enable this on)? Can you estimate if there will be a performance penalty in this scenario?


Yes, I think one option that applies to all supported interfaces is the right default. But I suggest keeping the interface list as an optional narrowing (empty = all).

Regarding the performance measurements, grilled Claude over the weekend and tested using openwrt rootfs docker images - the linear form (this PR as submitted, one rule per interface) scales with interface count. In my test environment, each rule costs about 13-15 ns per forwarded packet even when it fails at ct state new, and about 33 ns per new connection (where it also runs the iifname compare). Invisible at 2-8 interfaces, measurable at 32 interfaces, bad at 128 interfaces.

Claude comes up a alternative approach: a vmap form.
```
ct state new iifname vmap { "wan" : jump pbr_symset_0x010000, "wg0" : jump pbr_symset_0x020000, … }
ct direction reply ct mark & 0x00ff0000 vmap { 0x010000 : jump pbr_mark_0x010000, 0x020000 : jump pbr_mark_0x020000, … }
```
Below is the performance comparison table (measured deltas vs. no feature at all, median of 5 interleaved reps)

| interfaces | linear: UDP 64B pps | linear: new conn/s | vmap: UDP 64B pps | vmap: new conn/s |
|---|---|---|---|---|
| 2 | −5 % | −2 % | −3 % | −3 % |
| 8 | −4 % | −5 % | −7 %* | −2 % |
| 32 | **−15 %** | −9 % | +1 % | −5 % |
| 128 | **−38 %** | **−26 %** | +1 % | −8 % |

Also attaching the full report.
[REPORT.md](https://github.com/user-attachments/files/32161939/REPORT.md)

tracyhatemice:
> 2. This seems to obsolete the "default ICMP interface" feature, should we remove that?

I don't have a real world use case for `icmp_interface` myself, but they seem to address different problems: `icmp_interface` routes locally generated ICMP traffic, whereas `symmetric_response` ensures inbound traffic (including echo requests) returns along the same path. Because their purposes don't fully overlap, I agree with @egc112 that we should keep `icmp_interface`.

egc112:
Not sure on what hardware you tested. But on a slow single core router there might be a more measurable performance penalty.

egc112:
Note that I think pbr should already be able to pin port-forward return traffic today, with an ordinary policy and no new code.

The reason it should work is that mangle_prerouting runs at priority mangle (-150), which is before the routing decision — ip_rcv_finish() only calls ip_route_input_noref() once the PREROUTING hooks have completed. So a mark set there is an input to the route lookup rather than a reroute after the fact. And the reply from a forwarded host already carries that host's own source address; nothing in the kernel re-derives it, so the only thing that needs correcting is the egress interface — which is precisely what a mark can do. The un-DNAT back to the uplink address happens in POSTROUTING, well after routing, and does not interfere.

For a web server on 192.168.1.50 reached through a port forward on wan2, this is the whole configuration:

config policy
     option name 'forward-return-wan2'
     option src_addr '192.168.1.50'
     option src_port '80'
     option proto 'tcp'
     option interface 'wan2'

prerouting is the default chain, so option chain can be omitted. Replies from that host on port 80 are marked for wan2's table and leave by the interface the connection arrived on.

Of course this needs hand crafted rules and does not cover UPnP-forwards, port ranges or helper-created expectations so for a true multiwan package the connection tracking approach is certainly valuable so perhaps integrate it with mwan4?

tracyhatemice:
> Note that I think pbr should already be able to pin port-forward return traffic today, with an ordinary policy and no new code.
> ...
> Of course this needs hand crafted rules and does not cover UPnP-forwards, port ranges or helper-created expectations so for a true multiwan package the connection tracking approach is certainly valuable so perhaps integrate it with mwan4?

My use case for symmetric routing is as follows:

An OpenWrt router serves as Site Lab’s gateway and acts as client, establishing tunnels to the gateways at Sites A, B, and C. Each remote site has its own LAN, and the router is also configured to make the Lab LAN accessible through these tunnels (also using VRFs as tunnels can have overlapped endpoint addresses).

With symmetric routing enabled, return traffic from the Lab LAN always leaves through the same tunnel it arrived on, ensuring reliable connectivity to Lab LAN and from each remote site’s LAN.

egc112:
> > Note that I think pbr should already be able to pin port-forward return traffic today, with an ordinary policy and no new code.
> > ...
> > Of course this needs hand crafted rules and does not cover UPnP-forwards, port ranges or helper-created expectations so for a true multiwan package the connection tracking approach is certainly valuable so perhaps integrate it with mwan4?
> 
> My use case for symmetric routing is as follows:
> 
> An OpenWrt router serves as Site Lab’s gateway and acts as client, establishing tunnels to the gateways at Sites A, B, and C. Each remote site has its own LAN, and the router is also configured to make the Lab LAN accessible through these tunnels (also using VRFs as tunnels can have overlapped endpoint addresses).
> 
> With symmetric routing enabled, return traffic from the Lab LAN always leaves through the same tunnel it arrived on, ensuring reliable connectivity to Lab LAN and from each remote site’s LAN.

Yes that is an excellent use case for a multiwan package like mwan4 which is in the works and already partly integrated with PBR

PR #167, opened 2026-08-30
pbr-1.2.3: pass ip(2) an argument vector instead of a shell string

egc112:
sh.ip() received an argument vector from all 12 try_ip callers, flattened it with join(' '), and handed the result to system() as a string -- so /bin/sh re-split it on whitespace. Any argument containing a space was silently wrong, and UCI-derived values reached the shell as text.

system() has accepted an array since long before the oldest ucode pbr supports (uc_system()'s UC_ARRAY branch is present in 85922056 / OpenWrt 25.12), where it is execvp()'d with no shell involved. Verified on the DL-WRX36 running ucode-2026.01.16~85922056 and on an x86 master build running ucode-2026.07.09~b885dd0f. This is unlike fs.popen(), which is string-only on both and stays as it is.

Ten system() calls must keep a shell: they need 2>/dev/null to suppress the expected failure of deleting an absent rule, and system() offers no fd redirection. Those get sh.quote() on every interpolated value instead.

Also normalise cfg.prefixlength. UCI declares it a plain string with no bound, so /etc/config/pbr content reached an ip rule command line verbatim; clamp it the way uplink_ip_rules_priority already is. listen_port, read from /etc/config/network, is covered by the quoting above.

Both sources are root-owned config, so this is robustness rather than a privilege boundary -- but a space or a quote in the wrong UCI field produced a failure far from its cause.

mocklib gains argvs(). commands() flattens an array call with join(' '), so ['a b','c'] and ['a','b','c'] record identically and it structurally cannot see the defect being fixed. Four existing tests are updated for the now-quoted command text; 04_uplink_down_no_failed_to_start rolled its own system() capture that assumed strings and dropped the array-form logger call.

Tests: 59/59. Both new tests confirmed to fail when their fix is reverted.
check-ucode-syntax.sh against the router's own 85922056 build: 17 files, 0 failed.

### DIFF
--- a/files/lib/pbr/config.uc
+++ b/files/lib/pbr/config.uc
@@ -145,6 +145,13 @@ function create_config(uci_mod, ubus_mod, pkg) {
 		if (!match('' + cfg.procd_boot_trigger_delay, /^[0-9]+$/)) cfg.procd_boot_trigger_delay = '5000';
 		if (int(cfg.procd_boot_trigger_delay) < 1000) cfg.procd_boot_trigger_delay = '1000';
 
+		// suppress_prefixlength is interpolated into an `ip rule` command line,
+		// so it must be a bare number. UCI declares it a string with no bound,
+		// which let arbitrary text through to the shell. 0-128 is the protocol
+		// maximum; `ip -4` rejects anything above 32 on its own.
+		if (!match('' + cfg.prefixlength, /^[0-9]+$/)) cfg.prefixlength = '1';
+		if (int(cfg.prefixlength) > 128) cfg.prefixlength = '128';
+
 		if (!match('' + cfg.uplink_ip_rules_priority, /^[0-9]+$/)) cfg.uplink_ip_rules_priority = '30000';
 		if (int(cfg.uplink_ip_rules_priority) < 99) cfg.uplink_ip_rules_priority = '99';
 		if (int(cfg.uplink_ip_rules_priority) > 32765) cfg.uplink_ip_rules_priority = '32765';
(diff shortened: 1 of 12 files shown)


egc112:
Co-pilot comment:

> Direct recommendation: LGTM — approve and merge once CI completes. Your fixes close the real gap I found, the added comments make the intent explicit, and the 2>/dev/null audit is clean.
> 
> Summary of what you fixed and what I checked
> 
>     output.uc: you added the explanatory comment about the logger now being called as an argv array and why sh.quote() was removed — good.
>     sys.uc & mocklib.uc: the explanatory comments you mentioned are present in the PR (sys.uc has the six-line comment above ip(), mocklib.argvs() is documented). That covers intent, version note, and the reason for the new test recorder.
>     2>/dev/null audit: you ran a quick audit and confirmed all eleven shell-retaining system() calls are quoted; you listed their locations (sys.uc:74; nft.uc:693,696,710,713; pbr.uc:1239,1242,1261,1262,1266,1267) — none were missed.
>     Coercion ('' + prio): your explanation is sound. uc_system() and join() both use the same ucv_to_string path on the pinned ucode, so explicit coercion is unnecessary and would be inconsistent.
>     Compatibility: you correctly distinguished parsing verification (ucode-syntax) from runtime verification, and you also ran hardware tests on the cited ucode/dev builds — that gives the needed runtime confidence.
>     Helper extraction / try_cmd(): I agree it’s a reasonable follow-up rather than part of this PR; you noted try_cmd() still flattens args and would be the right target later.
> 
> Risk & readiness
> 
>     Risk: Low. Changes are conservative where a shell is required and otherwise fix argument-splitting/interpretation bugs.
>     Tests: New/updated tests exercise the change and the author reports 59/59 passing locally; make sure CI runs them on the merge branch.
>     Merge readiness: Ready to merge after CI passes. No blocking code issues remain.

PR #168, opened 2026-08-31
pbr-1.2.3: remove only pbr's own entry from dnsmasq's addnmount list

egc112:
resolver._dnsmasq_instance_cleanup() runs on every pbr stop and restart, and
called ctx.delete('dhcp', instance, 'addnmount', pkg.dnsmasq_file). ucode's
uci.delete() is delete(config, section[, option]): uc_uci_delete() reads three
arguments and a fourth is silently ignored, so with ptr.option already set
libuci deleted the entire option -- every other package's mount with it -- and
still returned true. The shell implementation this was converted from used
uci_remove_list (uci del_list); the ucode port lost that. list_remove() is the
call that removes a single value, and it is present in the oldest ucode pbr
supports.

Measured on a DL-WRX36 running ucode-2026.01.16~85922056, against a scratch
config dir:

  before          : [ "/tmp/adblock-backup", "/var/run/pbr.dnsmasq", "/var/run/https-dns-proxy" ]
  delete() rv     : true
  after delete()  : null
  list_remove rv  : true
  after list_rm() : [ "/tmp/adblock-backup", "/var/run/https-dns-proxy" ]

The adjacent ctx.reorder('dhcp', instance, idx) is dropped too. reorder(config,
section, index) moves a section within the package, so pbr was shuffling the
dnsmasq section around /etc/config/dhcp using a list index that has nothing to
do with section order.

This is a hard DNS outage for anyone holding a second addnmount, not a
degradation. Confirmed end to end on the same router with adblock 4.5.7-r3 and
adb_dnsshift=1, where adblock's conf-dir entry is a symlink into
/tmp/adblock-backup: one pbr restart drops that mount, and dnsmasq cannot read
its own config.

  addnmount before: '/var/run/pbr.dnsmasq' '/tmp/adblock-backup'
  addnmount after : '/var/run/pbr.dnsmasq'
  daemon.crit dnsmasq[1]: cannot access /tmp/dnsmasq.cfg01411c.d/adb_list.overall: No such file or directory
  daemon.crit dnsmasq[1]: FAILED to start up          (procd respawn, every 5s)
  nslookup openwrt.org 127.0.0.1 -> connection timed out; no servers could be reached

Reloading adblock re-adds its addnmount, which is why that appears to fix it --
until the next pbr restart. pbr's configure() re-adds pbr's own entry, so pbr
itself keeps working and only the other package's directory goes missing.

Exposure is narrower than it looks: adblock adds an addnmount only when
adb_dnsshift=1 (default 0, adblock.sh:37); with shift off it writes the list
straight into dnsmasq's conf-dir, which procd already mounts, and f_extconf()
actively del_lists the entry. The victims are shift=1 users and anything else
holding a second addnmount.

With the fix deployed on that router, four consecutive pbr restarts kept
adblock's entry and both jail mounts, with no duplicates and no dnsmasq
failures. pbr's own entry moves to the end of the list each cycle -- removed,
then re-appended -- which dnsmasq does not care about.

list_remove() is section-scoped exactly as delete() was, verified with two
dnsmasq sections and both addressing forms pbr uses ('guest' and '@dnsmasq[1]').
Worth noting the old behaviour was worse with several instances: configure()
cleans every instance before setting up the ones resolver_instance names, so
naming one instance stripped the list off the instances pbr was told to leave
alone, and those never got anything re-added.

mocklib's uci cursor gains list_remove(). It is array-only, and delete() keeps
ignoring a fourth argument, both matching the C implementation -- otherwise a
mock more helpful than libuci would hide this class of defect. A plain-string
option is left untouched by list_remove(), which also returns true.

Tests: 60/60. The new test is confirmed to fail 5 of its 6 checks when the fix
is reverted. check-ucode-syntax.sh against the router's own 85922056 build:
17 files, 0 failed.

### DIFF
--- a/files/lib/pbr/nft.uc
+++ b/files/lib/pbr/nft.uc
@@ -848,12 +848,16 @@ function create_nft(fs_mod, config, sh, output, pkg, platform, network, V, state
 		if (!ctx.get('dhcp', instance)) return false;
 		let confdir = resolver._dnsmasq_get_confdir(instance);
 		if (confdir) unlink(confdir + '/' + pkg.name);
+		// Drop ONLY our own entry. uci.delete() takes three arguments
+		// (config, section, option) -- a fourth is silently ignored, so
+		// delete(..., 'addnmount', file) wipes the whole list, taking every
+		// other package's mount (adblock's, most visibly) with it, and their
+		// directories stop being mounted into dnsmasq's jail on its next
+		// start. list_remove() is the call that removes a single value.
 		let current = ctx.get('dhcp', instance, 'addnmount');
 		if (type(current) == 'array') {
-			let idx = index(current, pkg.dnsmasq_file);
-			if (idx >= 0) {
-				ctx.reorder('dhcp', instance, idx);
-				ctx.delete('dhcp', instance, 'addnmount', pkg.dnsmasq_file);
+			if (index(current, pkg.dnsmasq_file) >= 0) {
+				ctx.list_remove('dhcp', instance, 'addnmount', pkg.dnsmasq_file);
 				ctx.save('dhcp');
 			}
 		}
--- a/tests/03_nft_rules/12_dnsmasq_addnmount_kept
+++ b/tests/03_nft_rules/12_dnsmasq_addnmount_kept
@@ -1,0 +1,110 @@
+Testing that dnsmasq cleanup removes only pbr's own addnmount entry.
+
+dnsmasq runs jailed, and every directory it must see inside that jail is listed
+as an `addnmount` on its uci section -- pbr's /var/run/pbr.dnsmasq, adblock's
+/tmp/adblock-backup, and whatever else a package added. The list is shared state.
+
+uci.delete() takes three arguments (config, section, option); a fourth is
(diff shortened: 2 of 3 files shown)


egc112:
co-pilot comment:

> 
> Overall Status: Ready to merge | Confidence: High
> 
> Positive Signals:
> 
>     Clear, well-documented root cause analysis
>     Minimal, surgical fix (only necessary changes)
>     Comprehensive test coverage with regression validation
>     End-to-end validation on real hardware (DL-WRX36 router)
>     All 60 existing tests pass
>     Syntax validation passes (17 files checked)
>     Mergeable state is clean with no conflicts
>     Member contributor (author_association: MEMBER)
> 
> No Blocking Issues:
> 
>     No CI failures noted
>     No review comments or requested changes
>     Code quality is high with appropriate inline documentation
>     Changes are well-scoped and do not introduce side effects
> 
> Note: The exposure of this bug is narrower than it appears — adblock only adds addnmount when adb_dnsshift=1 (default 0), but for anyone with that setting enabled or other packages mounting into dnsmasq's jail, this was causing hard DNS outages on every pbr restart.

PR #170, opened 2026-08-31
pbr-1.2.3: keep a policy name from breaking the whole nft ruleset

egc112:
A policy name is interpolated raw into every nft comment the policy produces: the rules in `pbr.uc`, the sets in `nft.uc`, and the trailing `# comment` on dnsmasq's nftset lines. An nft comment is a double-quoted string with **no escape for the quote** and a hard **128-byte** limit, and `nft_file.apply()` validates the generated file with a single `nft -c -f`.

So a policy name is not a cosmetic field. One holding a `"`, or one that simply runs long, makes nft reject the file, pbr installs nothing, and **all** policy routing stops — the same "one policy kills everything" class as the negated-domain bug in #159. The symptom is `Installing fw4 nft file [x]`.

Both are reachable from the WebUI, and the length one is the likelier: nobody types a quote into a policy name by accident, but a descriptive name reaches 128 bytes easily.

## What nft actually accepts

Verified against nftables 1.0.9:

| in the name | result |
|---|---|
| `"` | `syntax error, unexpected string, expecting newline or semicolon` |
| >128 **bytes** | `Error: comment too long, 128 characters maximum allowed` |
| `'` `;` `\` `{}` `$`, literal newline | parse fine |
| incomplete UTF-8 sequence at the cut | accepted |

The limit counts bytes, not characters, so 65 two-byte characters (130 bytes) already fail while 64 pass — which is why the truncation is a byte truncation.

## The change

A new `V.nft_comment()` folds `"` to `'`, folds CR/LF/TAB to a space, and truncates to 128 bytes. The quote becomes an apostrophe rather than being dropped so `My "work" laptop` still reads the way its author wrote it.

It is applied once in `policy_process()` and `dns_policy_process()`, where the name enters, rather than at each of the dozen interpolations downstream. That covers all three carriers at once, and the error and log lines then quote exactly what ends up in the ruleset.

The newline is folded for a second reason: the same string is written as a trailing `# comment` on a line of dnsmasq's nftset config, where an embedded newline injects a config line of its own. That failure mode is not visible to nft at all.

## Testing

`tests/03_nft_rules/13_policy_name_breaks_ruleset`, three policies (a quote, a 251-byte name, an embedded newline), 9 checks. It scores **3/9 on the unfixed code**, so it is not vacuous.

Beyond the mocked assertions, the ruleset pbr actually generates for those three policies was fed to a real `nft -c -f`:

```
unfixed: Error: syntax error, unexpected string, expecting ... semicolon
         Error: comment too long, 128 characters maximum allowed   (x3)
fixed:   (no parse or comment errors)
```

- `Ran 61 tests, 61 okay, 0 failures`
- ucode syntax gate against the pinned `85922056` (OpenWrt 25.12) build: `checked 17 file(s), 0 failed`; the helper's runtime behaviour was also exercised on that interpreter, since byte-truncation and `replace()` semantics are what the fix rests on.
- No new message codes, so compat stays at 36.

## Not in this change

`config.uc` builds `_nft_set_params` with `gc_interval "<val>"` and `timeout "<val>"` from UCI. Chasing that turned up something worse than an escaping gap, so it is being handled as its own change rather than folded in here: the emitted syntax is invalid regardless of the value, because the shell version's `"$var"` quoting was carried into the ucode port as literal quotes in the nft text, `gc_interval` should be `gc-interval`, and the port also emits `timeout` twice. Any non-empty `nft_set_timeout` or `nft_set_gc_interval` therefore makes nft reject the whole file. Confirmed on nftables 1.1.6 on a 25.12 router. Defaults are empty, so a default install is unaffected.

### DIFF
--- a/files/lib/pbr/pbr.uc
+++ b/files/lib/pbr/pbr.uc
@@ -636,6 +636,12 @@ function create_pbr(fs_mod, uci_mod, ubus_mod) {
 	
 	function dns_policy_process(uid, enabled, name, src_addr, dest_dns, dest_dns_port) {
 		if (enabled != '1') return 0;
+		// The name reaches nft raw -- as the comment on every rule the policy
+		// emits, on the sets it creates, and as the trailing comment on
+		// dnsmasq's nftset lines. Sanitise it here, where it enters, rather
+		// than at each of the dozen interpolations downstream; the errors and
+		// log lines below then quote exactly what ends up in the ruleset.
+		name = V.nft_comment(name);
 	
 		src_addr = replace(src_addr, /[,;{};]/g, ' ');
 		dest_dns = replace(dest_dns, /[,;{}]/g, ' ');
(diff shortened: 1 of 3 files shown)


egc112:
Co-Pilot comment:
Assessment: This PR is ready to merge. It's low-risk, well-tested (suite + hardware), properly scoped, and directly addresses a confirmed production issue. The code is clear, comments are defensive, and no feedback requires action.

PR #171, opened 2026-08-31
pbr-1.2.3: emit nft set time values in a form nft accepts

egc112:
Any non-empty `nft_set_timeout` or `nft_set_gc_interval` made nft reject the generated file, so pbr installed nothing and **all** policy routing stopped. This is not an escaping gap — the syntax was invalid whatever the value was.

Found while writing the "not in this change" note on #170.

## Three defects, each fatal on its own

**The values were emitted in double quotes.** Those quotes were the old shell init script's own — `gc_interval "$nft_set_gc_interval";` — where the shell consumed them before nft ever saw the line. The ucode port copied them across as literal characters, and nft answers `unexpected quoted string, expecting string`. `policy` was ported correctly, which is why only these two lines change.

**`gc_interval` is not an nft keyword** — it is `gc-interval`. Wrong in the shell version too, so that option has never worked in any release, while pbr's own README documents it as `gc-interval`.

**`timeout` was emitted twice** — once folded into the flags string by the port, once from `set_parts`. The shell version's flags string was flags only.

Here is what r93 actually generates with `nft_set_flags_timeout=1`, `nft_set_timeout=6h`, `nft_set_gc_interval=5m`:

```
add set inet fw4 pbr_wan_4_dst_ip_cfg0d6ff5 { type ipv4_addr; auto-merge; counter;
  flags interval, timeout; timeout 6h; gc_interval "5m"; policy performance; timeout "6h";  comment "ipleak";}
```

## Validation

An nft time value is a run of `<decimal><unit>` with the unit one of `ms s m h d`. A bare number is rejected, and so is `w` for weeks. UCI declares both options free strings with no bound, so `1w` or `60` — neither unreasonable to type — reached the ruleset and killed it.

An unusable value is now dropped rather than emitted. These are tuning knobs; an ignored timeout beats no policy routing at all, and it is the same treatment `prefixlength` already gets in `load()`, so there is no new message code and compat is unchanged. The check also closes the injection route that unquoted values would otherwise open.

| value | nft 1.0.9 | nft 1.1.6 | validator |
|---|---|---|---|
| `1h`, `6h`, `1h30m`, `1h30m10s`, `2d`, `100ms`, `0s` | accepts | accepts | accepts |
| `1w`, `60`, `0`, `1.5h`, `1 h`, `-5m`, `abc` | rejects | rejects | rejects |

## Testing

`tests/02_config/09_nft_set_time_values` (8 checks, valid values) and `10_nft_set_time_values_rejected` (6 checks, `1w` plus an injection payload). They score **3/8 and 4/6 on the unfixed code**.

- `Ran 62 tests, 62 okay, 0 failures`
- ucode syntax gate against the pinned `85922056` (25.12) build: `checked 17 file(s), 0 failed`; the validator regex was also exercised on that interpreter value by value
- pbr's generated ruleset goes from rejected to accepted by a real `nft -c -f`

**On hardware** (x86 OpenWrt 25.12, nftables 1.1.6, pbr 1.2.3-r93, 90 policies). Stock r93 with `flags_timeout=1 / timeout=6h / gc_interval=5m` logs `pbr 1.2.3-r93 FAILED TO START!!!`. With this change the same config starts clean and the kernel reports:

```
set pbr_wan_4_dst_ip_cfg0d6ff5 {
        flags interval,timeout
        timeout 6h
        gc-interval 5m
        elements = { 3.173.21.63 ... expires 5h59m57s530ms, ... }
```

With `nft_set_timeout=1w` and an injection payload in `nft_set_gc_interval`, both are dropped, `flags interval, timeout;` is kept, the ruleset is accepted and no payload appears anywhere in the file.

## Two things found alongside, deliberately not fixed here

**Set attributes cannot change on a running system.** pbr's named sets survive `service pbr stop` — the include file is removed and all 32 `pbr_*` sets are still in `inet fw4` — so the next start emits an `add set` whose spec differs from the live one and the pre-check fails with `Error: Could not process rule: File exists`. Enabling these options therefore needs a `service firewall restart` (which does clear them) or a reboot; a plain `service pbr restart` can never apply them. This affects any set-attribute change, not just these options.

**An element's timeout is never refreshed.** Measured on a live domain set: after forcing a dnsmasq cache miss and a genuine re-add, `expires` keeps counting down from the original insertion. So `nft_set_timeout=6h` drops an address 6h after it first appeared however busy it is, and it only returns on the next lookup after that. Worth a docs note before this option is recommended for the "sets only grow" problem.

### DIFF
--- a/files/lib/pbr/config.uc
+++ b/files/lib/pbr/config.uc
@@ -156,16 +156,32 @@ function create_config(uci_mod, ubus_mod, pkg) {
 		if (int(cfg.uplink_ip_rules_priority) < 99) cfg.uplink_ip_rules_priority = '99';
 		if (int(cfg.uplink_ip_rules_priority) > 32765) cfg.uplink_ip_rules_priority = '32765';
 
-		// Build nft_set_flags string
+		// An nft time value is a run of <decimal><unit>, unit being ms, s, m, h
+		// or d -- `1h`, `6h`, `1h30m`, `100ms`. A bare number is NOT accepted,
+		// and neither is a `w` for weeks; both were checked against nftables
+		// 1.0.9 and the 1.1.6 in OpenWrt 25.12. UCI declares these two options
+		// a free string with no bound, so a typo went straight into the set
+		// declaration. Drop a value nft cannot use rather than emit it: these
+		// are tuning knobs, and a whole ruleset nft refuses -- which stops ALL
+		// policy routing -- is a far worse outcome than an ignored timeout.
+		// Same treatment as prefixlength above.
+		if (cfg.nft_set_timeout && !match(cfg.nft_set_timeout, /^([0-9]+(ms|s|m|h|d))+$/))
+			cfg.nft_set_timeout = '';
+		if (cfg.nft_set_gc_interval && !match(cfg.nft_set_gc_interval, /^([0-9]+(ms|s|m|h|d))+$/))
+			cfg.nft_set_gc_interval = '';
+
+		// Build nft_set_flags string. Flags ONLY: the set's default timeout is
+		// emitted once, below. Appending it here as well put `timeout` in the
+		// declaration twice.
 		let nft_set_flags = '';
 		let fi = cfg.nft_set_flags_interval;
 		let ft = cfg.nft_set_flags_timeout;
 		if (fi && ft) {
-			nft_set_flags = 'flags interval, timeout' + (cfg.nft_set_timeout ? '; timeout ' + cfg.nft_set_timeout : '');
+			nft_set_flags = 'flags interval, timeout';
 		} else if (fi && !ft) {
 			nft_set_flags = 'flags interval';
 		} else if (!fi && ft) {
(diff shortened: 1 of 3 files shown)


egc112:
Co-pilot review:

> Strengths:
> 
>     Clear, well-documented bug fixes addressing three distinct defects
>     Excellent test coverage with edge cases (valid values, invalid values, injection attempts)
>     Thorough PR description explaining the history and nft syntax requirements
>     Hardware validation confirms the fix resolves the production issue
>     Low code complexity — straightforward string matching and keyword corrections
>     No breaking changes (dropped invalid values gracefully, same behavior as existing prefixlength handling)
>     Regex validation is appropriate and well-scoped
> 
> Status: Ready for merge
> 
>     No blocking issues identified
>     Clean mergeable state
>     All tests passing
>     Code follows the project's existing patterns (e.g., same validation approach as prefixlength)
> 
> Note: The PR author identifies two additional issues found during investigation but deliberately omits fixes for them (related to set attribute persistence and element timeout refresh). This is good engineering discipline — keeping the scope focused.

PR #172, opened 2026-08-31
pbr-1.2.3: quote every argument sh.try_cmd() puts on a command line

egc112:
`sh.try_cmd()` was the last place in `sys.uc` building a shell command line out of caller-supplied words without quoting them — the follow-up left open by #167.

Its 22 call sites are all `ip route replace` and `ip rule add`, carrying gateway addresses, device names and table ids. A value containing a space was re-split into two arguments; a value containing a shell metacharacter was interpreted.

## Verified on hardware, not just in the suite

x86 OpenWrt 25.12, `ucode 2026.01.16~85922056` — the same SHA the CI syntax gate pins — calling `try_cmd` directly with a real shell:

```ucode
try_cmd(e, '/bin/touch', '/tmp/q/one two')
try_cmd(e, '/bin/touch', '/tmp/q/x;touch /tmp/q/PWNED')
```

| | stock r93 | this branch |
|---|---|---|
| space-bearing argument | re-split into `one` and `two` | one file, `one two` |
| injection payload | **executed — `PWNED` created** | not executed |
| files left behind | `PWNED\|one\|two\|x\|` | `one two\|` |

So this is reachable on shipped hardware, not only in theory.

Separately, the patched `sys.uc` was deployed to that router and `pbr` restarted: it started clean, the routes `try_cmd` installs came out byte-identical to baseline (`default via 192.168.0.1 dev eth0`, `default dev wg_ch scope link`, …) and no `errorTryFailed` appeared. The quoting is transparent to real `ip`.

## Why not an argument array, like ip()

`try_cmd` goes through `run()`, which appends `>/dev/null 2>&1`, and `system()` has no fd redirection. The suppression is deliberate rather than incidental: every caller is a command whose failure is routine — a route replace on an interface that has no gateway yet — which `pbr` records as `errorTryFailed` and carries on from. Converting to an array would push those routine failures into the log.

So the shell stays and every element is `quote()`d instead, which is the rule #167 set for the calls it could not convert.

The recorded error keeps the **unquoted** join. That string is rendered as `Command failed: %s` in `service pbr status`; it is displayed, not executed, and shell quoting there would be noise the user never typed.

No call site changes — `try_cmd` takes a rest parameter, so the quoting is applied once inside it. `map()` was confirmed present on the router's own interpreter, since it is the one function the fix introduces.

## Scope

The other unquoted `sh.run()` interpolations were checked and deliberately left: `tid` comes from a `(\d+)` capture against `/etc/iproute2/rt_tables`, `priority` is computed arithmetic, `dryrun_dir` is a constant. None is a UCI-derived string, so none is this bug class.

## mocklib.fail_command()

Added because nothing could reach a failed-command path before — the `system()` stub always returned 0, so the unquoted `info` had no way of being pinned. It makes the stub report a non-zero exit for commands matching a needle, with the same needle semantics as `commands()`; `clear_commands()` resets it, and tests that do not opt in are unaffected. Same spirit as the `argvs()` recorder added in #167.

## Testing

- `tests/01_validation/06_sh_try_cmd_quoted`, 15 checks — **8/15 fail on the unfixed code**. Includes single-quote and repeated-single-quote values, which pin that `quote()`'s `'\''` escaping holds: that depends on `replace()` with a string needle being global in ucode. Checked against a real `/bin/sh` — `'O'\''Connor'` round-trips to `O'Connor`, and `'a'\''b'\''c'` to `a'b'c`
- `Ran 64 tests, 64 okay, 0 failures`
- ucode syntax gate against the pinned `85922056` build: `checked 17 file(s), 0 failed`
- no new message codes, so compat stays at 36

### DIFF
--- a/files/lib/pbr/sys.uc
+++ b/files/lib/pbr/sys.uc
@@ -81,10 +81,17 @@ function create_sys(fs_mod, pkg) {
 		return system([pkg.ip_full, ...args]);
 	}
 
+	// Callers hand this an argument VECTOR too, but unlike ip() it cannot go to
+	// system() as an array: every caller is a command whose failure is expected
+	// and routine -- a route replace on an interface that has no gateway yet --
+	// so its stderr must be suppressed, and system() has no fd redirection (see
+	// run()). The shell therefore stays, and every element is quote()d instead,
+	// which is what keeps a value carrying a space or a shell metacharacter as
+	// one argv word. The recorded error carries the UNQUOTED join: that string
+	// is shown to the user, not executed.
 	function try_cmd(errors, ...args) {
-		let cmd = join(' ', args);
-		if (run(cmd) != 0) {
-			push(errors, { code: 'errorTryFailed', info: cmd });
+		if (run(join(' ', map(args, quote))) != 0) {
+			push(errors, { code: 'errorTryFailed', info: join(' ', args) });
 			return false;
 		}
 		return true;
--- a/tests/01_validation/06_sh_try_cmd_quoted
+++ b/tests/01_validation/06_sh_try_cmd_quoted
@@ -1,0 +1,115 @@
+sh.try_cmd() is the last place in sys.uc that builds a shell command line out
+of caller-supplied words. Unlike sh.ip(), it cannot hand system() an argument
+array: run() suppresses the command's output with `>/dev/null 2>&1`, system()
+has no fd redirection, and the suppression is not incidental -- every caller is
+an `ip route replace` whose failure is routine, on an interface that has no
+gateway yet.
+
+So the shell stays and each element is quote()d instead. That is the same rule
+#167 applied to the calls it could not convert: a value carrying a space stays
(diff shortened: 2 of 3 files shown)


egc112:
Copilot review:
Thanks — that update looks good.

Summary of the maintainer's reply

    Adopted: single-quote test case (it found a real coverage gap) and the mocklib note about quoted command matching.
    Declined: capturing the quoted command in a local variable, adding a compatibility note about map(), and trimming the explanatory comment (kept as house style).
    Tests: suite 64/64; ucode syntax gate clean against the pinned build.

Recommended next step

    Approve & merge once CI finishes (no further code changes necessary).

Paste-ready approval comment you can post on the PR Thanks — I reviewed the follow-up. The single-quote test and mocklib documentation look good and exercise the edge cases you documented; the other suggestions you declined make sense given the explanation and existing coverage. The test suite and the ucode syntax gate are clean. I’m happy to approve this for merge once CI is green.

PR #173, opened 2026-08-31
pbr-1.2.3: warn once when an interface has no discoverable gateway

egc112:
A reporter running 1.2.3-r93 posted a clean start that carried the same line twice:

```
WARNING: Unknown IPv6 gateway for 'interface:wan6; device:eth1 '.
WARNING: Unknown IPv6 gateway for 'interface:wan6; device:eth1 '.
WARNING: Unknown IPv6 gateway for 'interface:wanb; device:lan1 '.
```

Cosmetic — routing is unaffected and the underlying report is correct — but the repeat reads as two separate faults, and the second line carries nothing to tell it apart from the first. The third line is a different interface and is right to be there.

## Cause

`get_gateway4()` and `get_gateway6()` each rewrite their interface argument to the uplink of their own address family before looking anything up: `get_gateway6()` maps the IPv4 uplink to `uplink_interface6`, `get_gateway4()` does the mirror image. The warning is then built from the *rewritten* name plus the device.

On the ordinary `wan` + `wan6` pair — one device, one upstream — processing `wan` produces a warning labelled `wan6`, and processing `wan6` produces the same one again. `state.warnings` had no dedupe, so both were recorded.

## A split uplink is worse off

With `uplink_interface` and `uplink_interface6` on *different* interfaces, the IPv4 uplink rewrites to `uplink_interface6` and picks up that interface's device (`pbr.uc:1122-1124`), landing on a line identical to the one the IPv6 uplink raises for itself — three warnings, two of them the same. Covered by the second test.

## What it dedupes on

Rendered text — `code` plus `info` — and nothing else. Not the device, not the underlying cause. So:

- two interfaces that genuinely have no gateway still warn separately **even when they share a device** — that is the reporter's `interface:wanb; device:lan1` line, and the `interface:wan` line a split uplink produces alongside `wan6` on the same `eth1`
- an interface whose device changed between calls still warns twice

Only an entry that would print as a byte-for-byte repeat is dropped. Both new tests assert this direction explicitly, and those assertions pass before and after the fix — they are guards against over-collapsing, not part of the reproduction.

## Verified on hardware, not just in the suite

x86 OpenWrt 25.12 (`r33164`), `ucode 2026.01.16~85922056` — the SHA the CI syntax gate pins — driving the real module with every lookup failing:

| case | stock r93 | this branch |
|---|---|---|
| default pair (`wan`+`wan6`, one device) | 2 warnings | **1** |
| IPv4 mirror | 2 warnings | **1** |
| distinct interfaces (`wan6`, `wanb`, `vpn`) | 4 warnings | **3** |
| same interface, different device | 2 warnings | 2 |

The IPv4 mirror is worth noting: `get_gateway4()` carries the same rewrite in reverse, so it had the same duplicate.

Separately, the patched `network.uc` was deployed to that box and `pbr` restarted. Against a pristine r93 run the generated `30-pbr.nft`, the IPv4 and IPv6 `ip rule` output and the `ubus` service data were all byte-identical, with `errors` and `warnings` empty in both. The dedupe is transparent wherever no duplicate exists. The file was restored to the packaged version afterwards.

## Scope

Consumers only iterate the array and test it for emptiness — `service_started` in `pbr.uc` and `status.js` in `luci-app-pbr` both — so no count changes meaning anywhere.

No message text and no message code changed, so this is not a catalog change and the compat stamp stays where it is.

The two new tests score 4/6 and 3/6 on the unfixed code.

### DIFF
--- a/files/lib/pbr/network.uc
+++ b/files/lib/pbr/network.uc
@@ -232,6 +232,18 @@ function create_network(fs_mod, config, sh, pkg, platform, V) {
 		return '';
 	}
 
+	// Record a warning unless an identical one is already there. Both gateway
+	// helpers below rewrite their interface argument to the uplink of their own
+	// address family, so on the usual wan + wan6 pair -- one device, one
+	// upstream -- processing 'wan' and then 'wan6' produces the same warning,
+	// with the same info string, twice. Two identical lines read as two faults;
+	// one interface with no gateway is one warning.
+	function push_warning_once(warnings, code, info) {
+		for (let w in warnings)
+			if (w.code == code && w.info == info) return;
+		push(warnings, { code: code, info: info });
+	}
+
 	function get_gateway4(iface, dev, warnings) {
 		if (is_uplink6(iface)) iface = cfg.uplink_interface4;
 		let gw = network_get_gateway(iface);
@@ -248,7 +260,7 @@ function create_network(fs_mod, config, sh, pkg, platform, V) {
 			}
 			// Raise warning if no gw and not point-to-point
 			if (!gw && warnings && !is_p2p(dev))
-				push(warnings, { code: 'warningInterfaceRoutingUnknownGateway4', info: 'interface:' + iface + '; device:' + dev + ' ' });
+				push_warning_once(warnings, 'warningInterfaceRoutingUnknownGateway4', 'interface:' + iface + '; device:' + dev + ' ');
 		}
 		return gw;
 	}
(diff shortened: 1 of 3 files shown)


egc112:
Copilot review:

> Merge Readiness
> Blocking Issues
> 
> None detected. The bug fix is correct, the refactor is well-structured, and hardware testing passed.
> Pre-merge Checklist
> 
>     Bug fix validated on hardware (OpenWrt 25.12)
>     No message text or catalog changes (compat stamp unchanged)
>     Config migrations in 90-pbr are safe
>     Test coverage notes (4/6, 3/6 on unfixed code)
>     CI syntax gate pins specific ucode SHA (85922056ef7a...)
>     Optional: Squash into single commit or add a refactor-phase tag to clarify scope
> 
> Risk Level
> 
> Low for the bug fix itself. Medium for the refactor scope — if anything breaks in the ucode logic, the service won't start at all. Mitigation: the shell init.d is now minimal and hard to get wrong; all logic is in structured modules.

PR #175, opened 2026-09-02
pbr-1.2.3: keep Tor policies from breaking the ruleset when tor is down

egc112:
A Tor policy on a box where torrc exists but the tor daemon is not running takes the **whole ruleset** down:

```
Installing fw4 nft file [x]
pbr 1.2.3-r93 FAILED TO START!!!
/var/run/pbr.nft:40:111-117: Error: syntax error, unexpected comment
add rule inet fw4 pbr_dstnat … udp dport 53 redirect to : comment "Tor-DNS-UDP"
                                                        ^^^^^^^
```

### Cause

`env.tor_dns_port` / `tor_traffic_port` start empty (`platform.uc:35-36`) and were only filled by `interface_process.tor('create')`, which `pbr.uc:2044` gated on `net.is_tor_running()`. The rule emitter at `pbr.uc:574` gates on `net.is_tor(iface)` — the policy's interface **name** — and `is_supported_interface()` called tor supported unconditionally through `is_tunnel()` (`network.uc:117`).

The two sites disagreed, so the emitter interpolated an empty port and nft rejected the entire file. pbr fails to start and every other policy stops routing with it.

This is not an exotic state:

- pbr is `START=20`, OpenWrt's tor is `START=50` — at boot the daemon is never up yet.
- LuCI offers `tor` in the interface dropdown on nothing more than `stat('/etc/tor/torrc')` (`network.uc:369`).
- `is_tor_running()` also gated the health check that verifies fw4's `dstnat` chain exists, so in this state pbr emitted dstnat rules without having checked for the chain.

### Fix

The ports live in torrc, not in the running daemon, so read them whenever tor is **configured**:

- **`is_tor_configured()`** — torrc readable and tor not ignored. Registration and the `dstnat` health check gate on this instead of `is_tor_running()`, so the rules are right while tor is starting and begin working the moment it comes up.
- **`is_supported_interface()` answers false for `tor` with no torrc**, so such a policy is rejected like any other unavailable interface rather than emitting a rule nft will refuse. Uses the existing `errorPolicyUnknownInterface` — no new message code, **compat stays at 36**, no companion luci PR.
- **A guard in the emitter** refuses to build a Tor rule without a port. It should now be unreachable; an empty port costs the whole ruleset, so it is not left to reachability arguments.

`is_tor_running()` is kept — still the honest answer for reporting — with a comment saying not to gate rule generation on it again.

### Tests

| | with fix | without |
|---|---|---|
| `05_interfaces/18_tor_ports_without_running_service` | 6/6 | **2/6** |
| `05_interfaces/19_tor_policy_without_torrc` | 5/5 | **1/5** |

Suite: `Ran 69 tests, 69 okay, 0 failures`.

### Verified on hardware

x86 25.12 r33164, nftables 1.1.6, pbr 1.2.3-r93, ucode `2026.01.16~85922056` — the oldest supported, and both modules compile under it. Patch applied to the router's own r93 files, so the delta is exactly this change.

| scenario | stock r93 | patched |
|---|---|---|
| torrc present, tor **not installed**, tor policy | FAILED TO START, `redirect to :` ×5 | starts clean, reports `tor/53->9153/80,443->9140`, all five rules carry the torrc ports in the live kernel ruleset, `nft -c -f` accepts |
| torrc removed, tor policy kept | same failure | starts clean, `ERROR: Policy 'Claude Tor Port Test' has an unknown interface!`, 0 redirect rules, 0 dstnat rules, the other 69 policy rules intact |

The torrc fixture uses 9153/9140 rather than Tor's defaults throughout, so a code path that silently fell back to 9053/9040 would fail rather than pass by coincidence.

Depends on #174 — before that merged, the test harness rewrote both Tor port helpers to read a capture group that does not exist, so none of this was testable.

### DIFF
--- a/files/lib/pbr/network.uc
+++ b/files/lib/pbr/network.uc
@@ -162,10 +162,22 @@ function create_network(fs_mod, config, sh, pkg, platform, V) {
 		return V.str_contains(cfg.lan_device, d);
 	}
 	function is_ignored_interface(iface) { return V.str_contains_word(cfg.ignored_interface, iface); }
-	function is_tor_running() {
+	// Tor is usable as a policy target as soon as it is CONFIGURED: the
+	// redirect ports come out of torrc, not out of the running daemon. pbr is
+	// START=20 and OpenWrt's tor is START=50, so at boot the service is never
+	// up yet -- gating the port lookup on it left interface_process.tor()
+	// unrun, the ports empty, and every Tor rule interpolating 'redirect to :',
+	// which nft rejects for the whole file.
+	function is_tor_configured() {
 		if (is_ignored_interface('tor')) return false;
 		let content = readfile(pkg.tor_config_file);
-		if (!content || content == '') return false;
+		return !!(content && content != '');
+	}
+	// Whether the daemon is actually up. Fine for reporting, but do NOT gate
+	// rule generation on it: the ports live in torrc, the emitter keys off the
+	// policy's interface name alone, and pbr starts before tor at boot.
+	function is_tor_running() {
+		if (!is_tor_configured()) return false;
 		let svc = config.ubus_call('service', 'list', { name: 'tor' });
 		if (!svc?.tor?.instances) return false;
 		for (let k in keys(svc.tor.instances)) {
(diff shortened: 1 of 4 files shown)


egc112:
Copilot review:

> Merge Readiness
> 
> Status: Ready to merge
> 
>     ✅ Clean merge into 1.2.3 branch (mergeable, no conflicts)
>     ✅ Fully backward compatible (existing message codes, compat level 36)
>     ✅ Comprehensive test suite passes (69/69)
>     ✅ Hardware verified on oldest supported platform
>     ✅ No breaking changes
>     ✅ Depends on PR #174 (already mentioned, likely merged)
>

PR #177, opened 2026-09-02
pbr-1.2.3: fix icmp_interface on IPv6, and report a 'proto' that cannot work

egc112:
Three faults that stop the service from starting, all reachable from the LuCI UI or a plain UCI typo without a warning. pbr validates the whole generated file in one `nft -c -f`, so any one of them takes **every** policy down, not just the one at fault.

## 1. `icmp_interface` emitted invalid nft for IPv6

IPv6 headers have no `protocol` field — it is `nexthdr` — and the IPv6 ICMP protocol is `icmpv6`. pbr emitted `ip6 protocol icmp`, and because the generated file is validated in one pass, setting **Default ICMP Interface** took every policy down with it:

```
/var/run/pbr.nft:32:34-41: Error: syntax error, unexpected string
add rule inet fw4 pbr_output ip6 protocol icmp counter goto pbr_mark_0x020000
pbr 1.2.3-r95: not active
```

Measured against nftables 1.1.6:

| rule | verdict |
|---|---|
| `ip protocol icmp` | accepted |
| `ip6 protocol icmp` | **REJECTED** ← what pbr emitted |
| `ip6 nexthdr icmpv6` | accepted |

The `ipv6_enabled` guard was never the problem — the rule was already gated on it — so it is unchanged.

## 2. `proto` is a port qualifier, and says so nowhere

It only ever reaches a rule as a prefix on `sport`/`dport`, and pbr's own defaulting confirms the intent (`pbr.uc`): with no proto set it becomes `tcp udp` when a port is present and `all` when it is not.

So a policy naming a protocol **without** a port has it silently dropped and routes *every* protocol; and a protocol that has no ports, **given** a port, emits `icmp dport { 5000-5100 }`, which nft refuses — taking the whole ruleset with it.

Neither is exotic. The LuCI dropdown is built from `/etc/protocols` (46 entries on stock 25.12) and only `tcp`, `udp`, `sctp`, `dccp`, `udplite` can carry a port, so **41 of the choices offered** land in one case or the other. ICMP is the sharpest: it cannot have a port at all, so there is no way to use it in a policy — and pbr's actual mechanism for it is the unrelated `icmp_interface`.

Handled by severity rather than uniformly:

| config | outcome |
|---|---|
| port-capable proto, no ports | **warn**, and say what to do — `0-65535` does express "all of this protocol" |
| portless proto, no ports | **warn** — it cannot work; points at the *Default ICMP Interface* setting |
| portless proto **with** a port | **reject the policy** |

The third cannot be a warning: by the time nft refuses the file the router has stopped routing. Rejecting the one policy leaves the rest of the ruleset intact.

## 3. An unknown `chain` names a chain that does not exist

```
add rule inet fw4 pbr_input ip daddr { 192.0.2.12 } counter goto pbr_mark_0x010000
Error: No such file or directory; did you mean chain 'pbr_output' in table inet 'fw4'?
pbr 1.2.3-r95: not active
```

pbr creates `pbr_dstnat` plus `chains_list` — `prerouting`, `forward`, `output` — and has since before 1.2.2, whose shell implementation carries the same three. The README nonetheless documented `input` and `postrouting` for years, so this was reachable by *following the documentation*; a plain typo does the same thing. Corrected in the docs by mossdef-org/docs.mossdef.org#36.

`dstnat` is not in `chains_list` but **is** created, and a policy naming it works today, so it stays accepted rather than breaking such a config.

## The rule applied

| config | outcome |
|---|---|
| port-capable proto, no ports | **warn** — `0-65535` does express "all of this protocol" |
| portless proto, no ports | **warn** — cannot work; points at *Default ICMP Interface* |
| portless proto **with** a port | **reject the policy** |
| unknown chain | **reject the policy** |

The test is whether the policy still produces a *valid* ruleset. An ignored proto does — too broad, but the router keeps routing, so the user can be told and left alone. The other two produce a file nft refuses, where a warning would annotate an outage rather than prevent one.

## Compat

Four new message codes, so **36 → 37** in all five places. The paired **mossdef-org/luci-app-pbr#48** carries the other two, the WebUI strings, and filters the dropdown to protocols that can actually work.

`.info` stays a single string: the procd emission writes it with `json_add_string`, so an array reaches the WebUI as `[ "name", "proto" ]`. There is a test asserting that.

## Tests

| test | with fix | without |
|---|---|---|
| `04_policies/09_proto_and_chain_validation` | 20/20 | **8/20** |
| `05_interfaces/20_icmp_interface_ipv6_syntax` | 7/7 | **5/7** |

Suite: `Ran 71 tests, 71 okay, 0 failures`.

`13_split_uplink_chain_name` matched the literal `protocol icmp`, so after this change it would have silently checked only the IPv4 half; it now matches both forms.

## Verified on hardware

25.12 (x86 r33164, nftables 1.1.6, ucode `2026.01.16~85922056`) with every shape configured at once — an `icmp_interface`, a proto-without-ports policy, an icmp-with-port policy and a `chain='input'` policy. **Each of those alone stops r95 from starting.** Patched:

```
pbr: active in dynamic routing tables mode
ip protocol icmp / ip6 nexthdr icmpv6      accepted by nft -c -f
icmp dport: 0    pbr_input: 0    offending policies: 0 rules    rest intact

WARNING: Policy 'br-lan' (udp): 'proto' is ignored without a port, so every
         protocol is routed; set the port range to '0-65535' to match all of it
ERROR:   Policy 'smoke port range' (icmp): this protocol cannot match a port;
         unset the port, otherwise nft rejects the whole ruleset
ERROR:   Policy 'smoke chain forward' (input): unknown chain; use 'prerouting',
         'forward' or 'output'
```

Paired with mossdef-org/luci-app-pbr#48. Docs: mossdef-org/docs.mossdef.org#36.


egc112:
Co-pilot review:

> Looks good — your follow-ups address the review points and I agree with the decisions and fixes. Quick summary and next steps:
> 
> What you resolved (verified)
> 
>     Adopted: added the inline comment explaining why "proto with no ports" is a warning but "portless proto with a port" is an error.
>     Adopted: added mixed-case tests (e.g. "UDP", "ICMP") and tightened the assertions so they are scoped to the exact policy.
>     Adopted: replaced the switch with a proto_with_ports set lookup (cleaner and easier to test).
>     Confirmed: the for-in iteration in ucode yields values (not indices) for this interpreter — the loop is correct.
>     Confirmed: no leftover compat=36 occurrences across files/tests/docs (you grep’d and explained the single unrelated test fixture).
>     Confirmed: mossdef-org/luci-app-pbr#48 covers the UI side (message keys, dropdown filtering, LuciCompat/rpcdCompat → 37).
>     Confirmed: keeping .info as a string is deliberate and tested (changing to structured objects would require coordinated changes in procd/ubus/UI).
>     Documenting/changelog: noted that docs will be updated in a separate PR (the repo intentionally keeps docs in docs.mossdef.org).
> 
> Remaining status and recommendation
> 
>     Tests and hardware verification exist and exercise the two bug cases. I see no remaining code-blocking issues in the diff.
>     Recommendation: approve and merge once CI passes and you coordinate the luci-app-pbr change (merge order: either merge the luci-app PR alongside or ensure downstream consumers are updated quickly). Also open the docs PR described so users see the behavioral change (proto semantics + compat bump).
>     Risk: low-to-moderate (narrow and tested), no blockers left in code.
>

PR #180, opened 2026-09-03
pbr.user.dnsprefetch: wait for service completion

pesa1234:
## Problem

pbr.user.dnsprefetch currently treats the nft ruleset file mtime as a signal that the service is ready, then waits a fixed 15 seconds. The file is copied before fw4 reload, resolver cleanup, and the subsequent dnsmasq restart/cache flush have completed.

On a GL-MT6000 with a large adblock configuration (dnsmasq reports 464,589 additional local addresses), the prefetch started while dnsmasq was still being replaced. All 106 parallel lookups then timed out, even though direct queries to 1.1.1.1, queries through 127.0.0.1, the PBR output mark, and the pbr_wan route were working after startup.

## Fix

- Capture the exact pbr ucode worker PID before detaching the prefetch process and log the PID being followed.
- Wait for that worker to exit, placing prefetch after fw4 reload, set cleanup, and the resolver restart/cache-flush request.
- Probe dnsmasq after the barrier so resolution waits until the daemon is serving again.
- Avoid a redundant second HUP on the normal PID path; PBR has already restarted dnsmasq or requested its cache flush.
- Keep the nft mtime fallback for a missing or invalid worker PID. This fallback retains its own HUP, waits one second for signal delivery, and then probes readiness.
- Keep counting NXDOMAIN lookups, but reserve the failure marker for timeouts, SERVFAIL, and other hard lookup failures.

Waiting for the captured PID avoids the ambiguity of searching the process table: it follows only the ucode worker that sourced this script, while the existing setsid plus redirected stdio keeps the service call non-blocking.

## Validation

- /bin/sh syntax check passes.
- The existing 71-test suite passes as a regression check; it does not directly exercise this user script.
- OpenWrt test package pbr 1.2.3-r101 builds and verifies successfully.
- Hardware validation on the affected MT6000 confirms the normal path logs and follows the exact worker PID. It resolved 105 domains plus one known local NXDOMAIN with no timeouts and the final success marker.
- Two reload requests submitted one second apart were serialized by the service on this router. Both prefetch runs followed their own worker PID and completed without timeout, fallback use, process leaks, or FIFO leaks.
- The missing-self branch was rechecked in an isolated shell and emits the expected warning immediately without spawning a child.

The PKG_RELEASE bump used for router testing is intentionally not included because release bumps on this branch are handled separately.

### DIFF
--- a/files/usr/share/pbr/pbr.user.dnsprefetch
+++ b/files/usr/share/pbr/pbr.user.dnsprefetch
@@ -38,9 +38,10 @@
 # wait here made "/etc/init.d/pbr reload" take 62s instead of 3s). So the first
 # pass records the pbr ucode worker's pid, re-executes this file in its own
 # session and returns immediately. The detached copy waits for that worker to
-# exit: only then have fw4 reload, set cleanup and the resolver restart/cache
-# flush all completed. Watching the nft file is too early because pbr copies it
-# before running fw4; a fixed delay raced dnsmasq on large configurations.
+# exit: only then have fw4 reload, set cleanup and pbr's resolver restart/cache
+# flush request all completed. dnsmasq may still be applying that request, so
+# its readiness is probed below. Watching the nft file is too early because pbr
+# copies it before running fw4; a fixed delay raced dnsmasq on large setups.
 #
 # The service sources this file, so it cannot discover its own path ($0 is the
 # sourcing shell). Update this if you install it somewhere else.
@@ -89,17 +90,17 @@ else
 			''|*[!0-9]*|0|1) return 2 ;;
 		esac
 
+		output "Waiting for $packageName run (pid $pbr_pid) to finish..."
 		while kill -0 "$pbr_pid" 2>/dev/null; do
 			[ "$timeout_pbr_run" -eq '0' ] && return 1
 			sleep '1' && timeout_pbr_run="$((timeout_pbr_run - 1))"
 		done
 	}
 
-	# Fallback for an explicitly forced child invocation, where there is no pbr
-	# worker pid to follow. Normal service starts never take this path.
-	#
-	# Modification time of the ruleset pbr installs, or the empty string before
-	# the first one exists.
+	# The mtime fallback is used only when there is no usable pbr worker pid to
+	# follow, such as an explicitly forced child invocation. Normal service starts
+	# never take this path. Return the ruleset's modification time, or the empty
+	# string before the first one exists.
 	nft_stamp() {
 		[ -f "$packageNftFile" ] || return 0
 		date -r "$packageNftFile" +%s 2>/dev/null
@@ -120,6 +121,16 @@ else
 			[ "$timeout_nft_file" -eq '0' ] && return 1
 			sleep '1' && timeout_nft_file="$((timeout_nft_file - 1))"
 		done
+	}
+
+	pbr_ruleset_fallback_finished() {
+		local settle='15'
+		local stamp_before
+
+		output "$packageName worker pid unavailable, using ruleset fallback..."
+		stamp_before="$(nft_stamp)"
+		nft_file_installed "$stamp_before" || return 1
+		sleep "$settle"
 	}
 
 	nft_ready() {
@@ -173,25 +184,30 @@ else
 		output="$(nslookup "$1" 127.0.0.1)" && { echo '0' > "$pipe_nslookup"; return; }
 		reason="$(printf '%s' "$output" | grep -Eo -m 1 'NXDOMAIN|SERVFAIL|timed out')" && \
 		output "$_WARNING_ Lookup failed for $1 ($reason)"
-		echo '1' > "$pipe_nslookup"
+		[ "$reason" = 'NXDOMAIN' ] && echo '2' > "$pipe_nslookup" || echo '1' > "$pipe_nslookup"
 	}
 
 	nslookup_tracker() {
-		local entries errors result
+		local entries errors hard_errors result
 
 		while read -r rc; do
 			entries="$((entries + 1))"
-			[ "$rc" -eq '1' ] && errors="$((errors + 1))"
+			[ "$rc" -ne '0' ] && errors="$((errors + 1))"
+			[ "$rc" -eq '1' ] && hard_errors="$((hard_errors + 1))"
 		done < "$pipe_nslookup"
 
-		[ "${errors:-0}" -eq '0' ] && result="$__OK__" || result="$__FAIL__"
+		# NXDOMAIN is a valid DNS reply and may be permanent or intentional. Keep
+		# counting it, but reserve the failure marker for timeouts, SERVFAIL and
+		# other cases where no usable answer was received.
+		[ "${hard_errors:-0}" -eq '0' ] && result="$__OK__" || result="$__FAIL__"
 		output "Finished resolving $entries domain names in policies (${errors:-0} failed) $result"
 	}
 
 	main() {
 		local msg_abort='domain names in policies not resolved'
-		local settle='15'
-		local stamp_before domain
+		local barrier_rc='2'
+		local used_fallback='0'
+		local domain
 		local rc='0'
 
 		# $resolverSetSupported is not available here; read the setting pbr
@@ -207,18 +223,24 @@ else
 		mkfifo "$pipe_nslookup"
 
 		if [ -n "$PBR_DNSPREFETCH_PARENT_PID" ]; then
-			pbr_run_finished "$PBR_DNSPREFETCH_PARENT_PID" || {
+			pbr_run_finished "$PBR_DNSPREFETCH_PARENT_PID"
+			barrier_rc="$?"
+		fi
+
+		case "$barrier_rc" in
+			0) ;;
+			2)
+				used_fallback='1'
+				pbr_ruleset_fallback_finished || {
+					output "Timed out waiting for $packageName to install its ruleset, $msg_abort $__FAIL__"
+					rc='1'
+				}
+				;;
+			*)
 				output "Timed out waiting for $packageName to finish its run, $msg_abort $__FAIL__"
 				rc='1'
-			}
-		else
-			stamp_before="$(nft_stamp)"
-			nft_file_installed "$stamp_before" || {
-				output "Timed out waiting for $packageName to install its ruleset, $msg_abort $__FAIL__"
-				rc='1'
-			}
-			[ "$rc" -eq '0' ] && sleep "$settle"
-		fi
+				;;
+		esac
 
 		[ "$rc" -eq '0' ] && [ ! -f "$packageDnsmasqFile" ] && {
 			output "File $packageDnsmasqFile not found, $msg_abort $__FAIL__"
@@ -234,31 +256,31 @@ else
 			rc='1'
 		}
 
-		[ "$rc" -eq '0' ] && {
-			# dnsmasq only writes to an nft set on an upstream reply, so a name
-			# still in its cache would be answered without filling the set.
-			# pbr has already cleared the cache by this point on both of its
-			# paths, but doing it here too costs nothing and keeps this correct
-			# whichever path ran, and whatever the timing. Only ever after
-			# dnsmasq_ready() -- see the note there.
+		# On the parent-pid path pbr has already restarted dnsmasq or sent the
+		# cache-flushing HUP before its worker exited. Repeating it would force a
+		# second expensive re-read of large hosts/adblock configurations. The
+		# fallback cannot make that guarantee, so retain the flush there. Wait one
+		# second before probing to ensure the first answer cannot race ahead of
+		# dnsmasq handling the signal.
+		[ "$rc" -eq '0' ] && [ "$used_fallback" -eq '1' ] && {
 			killall -q -s HUP dnsmasq
-			# On large hosts/adblock configurations SIGHUP itself can take longer
-			# than the old one-second delay. A reply proves the final daemon has
-			# completed the cache flush and returned to its event loop.
-			if ! dnsmasq_ready; then
+			sleep '1'
+			! dnsmasq_ready && {
 				output "Dnsmasq is not answering after cache flush, $msg_abort $__FAIL__"
 				rc='1'
-			else
-				nslookup_tracker & exec 3>"$pipe_nslookup"
-				(
-					output "Resolving domain names in policies..."
-					while IFS='/' read -r _ domain _; do
-						[ -n "$domain" ] && run_nslookup "$domain" &
-					done < "$packageDnsmasqFile"
-					wait
-				)
-				exec 3>&-
-			fi
+			}
+		}
+
+		[ "$rc" -eq '0' ] && {
+			nslookup_tracker & exec 3>"$pipe_nslookup"
+			(
+				output "Resolving domain names in policies..."
+				while IFS='/' read -r _ domain _; do
+					[ -n "$domain" ] && run_nslookup "$domain" &
+				done < "$packageDnsmasqFile"
+				wait
+			)
+			exec 3>&-
 		}
 
 		rm -f "$pipe_nslookup"
--- a/files/usr/share/pbr/pbr.user.dnsprefetch
+++ b/files/usr/share/pbr/pbr.user.dnsprefetch
@@ -204,9 +204,12 @@ else
 		done < "$pipe_nslookup"
 
 		# NXDOMAIN is a valid DNS reply and may be permanent or intentional. Keep
-		# counting it, but reserve the failure marker for timeouts, SERVFAIL and
-		# other cases where no usable answer was received.
-		[ "${hard_errors:-0}" -eq '0' ] && result="$__OK__" || result="$__FAIL__"
+		# counting it without failing a partly successful run. Timeouts, SERVFAIL
+		# and other hard errors fail immediately; an all-NXDOMAIN run also fails
+		# because it did not resolve a single domain into a set.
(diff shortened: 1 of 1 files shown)


egc112:
Thanks for the detailed report and for testing this on hardware with a config
  that actually stresses it. I went through the change and checked its premises
  against the source; the design is right and the barrier is an improvement over
  what's on `1.2.3` today. A few things to fix before this goes in.

  ## What I verified

  - `$PPID` really is the pbr ucode worker. `sh.run()` is
    `system(cmd + ' >/dev/null 2>&1')` (`files/lib/pbr/sys.uc:27`) — a string, so
    ucode fork/execs `/bin/sh -c` directly and the wrapper sources the user file
    in that same shell. Confirmed by experiment: the sourcing shell's parent
    `comm` is `ucode`. busybox ash does set `$PPID`. 
  - Waiting for the worker to exit is strictly better than the mtime + settle it
    replaces. `pbr.uc:2120-2157` runs user files, then `nft_file.apply('main')`,
    `orphan_sets`, `flush_changed`, and finally `resolver.restart()` /
    `flush_cache()` — all inside that process, so its exit covers everything the
    15s settle was guessing at. `service_started()` is a second ucode invocation
    but it only reports; it touches no sets (`pbr.uc:2258`).
  - No deadlock risk: the worker never waits on the setsid child.

  So the barrier itself is sound. My concerns are about what happens after it.

  ## 1. Please keep the `sleep 1` before the post-HUP probe

  `killall` returns when the signal is *delivered*, not when it's handled.
  dnsmasq answers from its event loop and processes SIGHUP on the next wakeup, so
  the first `nslookup` can be answered *before* the flush begins — after which we
  launch every lookup into a daemon that then disappears for several seconds.
  That is precisely the failure you're fixing, so the probe shouldn't be able to
  race past it:

  ```sh
  killall -q -s HUP dnsmasq
  sleep '1'
  if ! dnsmasq_ready; then
  ```

  One second, in a detached copy nothing is waiting on.

  ## 2. Our own HUP is now redundant — and on your box it's the expensive part

  `resolver.flush_cache()` is literally `sh.run('killall -q -s HUP dnsmasq')`
  (`nft.uc:916-931`), and `resolver.restart()` leaves an empty cache by
  construction. Both run at the very end of the worker the child now waits for.
  So on the parent-pid path this script forces a *second* full re-read of those
  464,589 addresses, on the one machine where that re-read is slow enough to
  matter. 

  The comment "doing it here too costs nothing and keeps this correct whichever
  path ran" was true under the old barrier, where the child could get ahead of
  pbr's flush. Under the new one it isn't. I'd rather see the HUP kept only in
  the fallback branch, where it's still needed. If you'd prefer to leave it in
  place, then item 1 is not optional.

  ## 3. `pbr_run_finished` returning 2 aborts instead of falling back

  The `''|*[!0-9]*|0|1` guard exists to detect a pid we can't follow — but the
  caller treats any non-zero as a timeout:

  ```sh
  pbr_run_finished "$PBR_DNSPREFETCH_PARENT_PID" || {
          output "Timed out waiting for $packageName to finish its run, ..."
          rc='1' 
  }       
  ```     
  So the one case the guard is there to catch logs a misleading message and skips
  the prefetch entirely, never reaching the stamp + settle path that was written
  for exactly that situation. It needs to distinguish them, e.g.

  ```sh
  pbr_run_finished "$PBR_DNSPREFETCH_PARENT_PID"
  case "$?" in
          0) ;;
          2) <fall through to the stamp path> ;;
          *) output "Timed out waiting for $packageName to finish its run, $msg_abort $__FAIL__"; rc='1' ;;
  esac    
  ```     

  which probably reads best with the stamp + settle branch pulled into its own
  function.

  ## 4. `[✗]` on a permanent NXDOMAIN — worth reconsidering

  Your own run reports failure because of `wt-eu02.net`. A domain that is
  permanently unresolvable in someone's policy would now make *every* prefetch
  log `[✗]` forever, when `(1 failed)` already conveys it and there's nothing pbr
  can do about an NXDOMAIN. I'd fail only on a timeout or SERVFAIL — an answer we
  didn't get, rather than an answer we don't like — or only when nothing resolved
  at all. Happy to be argued out of it, but as written it changes the steady-state
  log line for a fair number of installs.

  ## 5. On the validation

  Nothing in the evidence distinguishes which of the two paths actually ran: if
  `$PPID` were empty the script falls back silently, and the improvement you
  measured (105/106, no timeouts) is fully explained by the post-HUP probe on its
  own. A log line naming the pid it's following would make that observable —
  otherwise, could you confirm from the logs which branch was taken? 

  Two other notes: the test suite doesn't cover this file (nothing in the repo
  tests or shellchecks it), so "71 passed" isn't evidence for this change. And
  since the barrier semantics changed, could you re-run the two-overlapping-reloads
  case and the missing-`_dnsprefetch_self` case? Those were the awkward ones last
  time round.

  ## 6. Nits

  - The header now says that on worker exit "fw4 reload, set cleanup and the
    resolver restart/cache flush [have] all completed". The restart has only been
    *issued* — the init script returns long before dnsmasq is serving, which is
    why `dnsmasq_ready()` exists. Worth rewording, since these comments are the
    documentation for this file.
  - The "Fallback for an explicitly forced child invocation" note ended up
    attached to `nft_stamp()`, leaving the old "Modification time of the ruleset"
    comment orphaned under a bare `#`. The `settle` local and the long
    `nft_file_installed()` comment now apply only to the fallback too.
   
  Not a blocker: `kill -0` can't distinguish the worker from a recycled pid, but
  we start polling milliseconds after it exists, so that window is negligible.

pesa1234:
Thanks for checking the premises and for the detailed review. I addressed the requested changes in b132d89:

1. The fallback HUP now keeps sleep 1 before its readiness probe.
2. The normal parent-PID path no longer sends a second HUP; only the mtime fallback does.
3. Return code 2 from pbr_run_finished now selects a dedicated pbr_ruleset_fallback_finished function. Only an actual wait timeout aborts with the timeout message.
4. NXDOMAIN is still counted and logged but does not select the failure marker. Timeouts, SERVFAIL, and other hard lookup failures do.
5. The normal path now logs the exact worker PID it follows; the fallback logs that it is being used. I also corrected the PR text to describe the 71 tests as regression coverage only, not coverage of this script.
6. The barrier comment now says the resolver request has completed and that dnsmasq may still be applying it. The fallback mtime and settle comments/locals are grouped with the fallback function.

I rechecked the missing-self branch in an isolated shell with a mock logger. It returned immediately and emitted:

    -t pbr [2] WARNING: /usr/share/pbr/pbr.user.dnsprefetch not found, domain names in policies not resolved [✗]

Shell syntax, diff checks, the 71-test regression suite, the r101 package build, and APK verification all pass. I have built r101 for the MT6000 and will post the observable PID-path and overlapping-reload hardware logs after that rerun.

pesa1234:
@egc112 

Hardware rerun completed on the affected GL-MT6000 with pbr 1.2.3-r101.

The observable normal start confirms the parent-PID branch was used:

    12:14:09 pbr [20029]: Waiting for pbr run (pid 19931) to finish...
    12:14:10 pbr: pbr 1.2.3-r101 started in dynamic routing tables mode.
    12:14:15 pbr [20029]: Resolving domain names in policies...
    12:14:16 pbr [20029]: Finished resolving 106 domain names in policies (1 failed) [✓]

There was no fallback message. The one counted failure was the known local NXDOMAIN for wt-eu02.net; there were no timeouts.

For the overlapping-reload case, two reload requests were submitted one second apart. Service processing serialized them on this router, but both detached prefetch runs followed their own worker PID and completed cleanly:

    pbr [20522]: Waiting for pbr run (pid 20405) to finish...
    pbr [20522]: Finished resolving 106 domain names in policies (1 failed) [✓]
    pbr [20926]: Waiting for pbr run (pid 20831) to finish...
    pbr [20926]: Finished resolving 106 domain names in policies (1 failed) [✓]

Again there were no timeouts, no fallback, and ls /var/run/pbr.nslookup.* returned no entries after completion, so there were no FIFO leaks.

The missing-self branch was also rerun separately in an isolated shell: it returned immediately, logged the expected missing-file warning, and did not spawn a child.

egc112:
Thanks — that was a fast and unusually thorough turnaround, and the observable
  hardware logs are exactly what I asked for. `pbr_ruleset_fallback_finished()`
  and the `barrier_rc` case read better than the sketch I gave you.

  I took the revision to a DL-WRX36 (adblock, ~204k domains) and stress-tested it
  against the released script. Summary: the barrier is right and I want it in.
  One correction that is mine to make, one remaining ask.

  ## 1. My item 2 was wrong — your change is right, its stated reason isn't

  I claimed the second HUP forces a re-read of the large hosts/adblock
  configuration. That is false, and you've now written it into the file:

  > Repeating it would force a second expensive re-read of large hosts/adblock configurations.

  Measured on hardware with **1,004,758 `local=` entries** loaded: SIGHUP costs
  **0.01s**. dnsmasq re-reads *hosts files* on SIGHUP, not its configuration, and
  adblock's dnsmasq backend is a conf-dir, so those entries are not re-read at
  all. The expensive operation is the restart — the init script returns after
  3.7s and dnsmasq only answers 9.8s after the restart began.

  So I checked whether dropping it is *safe*, which is the question that actually
  matters. Warm dnsmasq's cache with every policy domain, flush the sets, then
  reload: if pbr's own flush had not been processed before the burst, the lookups
  would be answered from cache and the sets would stay empty. **5/5 runs filled
  all 43 elements.** Your change stands; only the justification needs rewording —
  it is redundant, not expensive.

  Related, and worth knowing for future validation: **`nslookup` returning 0 does
  not prove the set was filled.** A cache hit answers perfectly well without
  dnsmasq ever writing to the nftset, so "105 of 106 resolved" cannot distinguish
  a working prefetch from one whose sets stayed empty. Element counts can. I
  checked it for this revision, so no action needed here.

  ## 2. The one thing I'd still like: validate `$PPID` before following it

  The new log line is what made this visible — good call adding it. Run by hand
  under a shell that is not pbr, `b132d89` logs:

  ```
  Waiting for pbr run (pid 10707) to finish...
  ```

  10707 was a throwaway `sh -c` shell, and there were **zero `cli.uc` processes
  alive** at that moment. It sat there until I killed that shell 46 seconds
  later. From a login shell that stays open it burns the whole 300s budget and
  then aborts with "Timed out waiting for pbr to finish its run" — and
  `pbr_ruleset_fallback_finished()`, which you just wrote for exactly this case,
  is unreachable, because `$PPID` is always numeric and alive. The numeric guard
  only catches 0/1/non-numeric, which in practice never happens.

  The check belongs in the parent, before detaching. It cannot go in the child:
  there, "no such pid" is ambiguous — it is the normal fast case (worker already
  exited, proceed) and the by-hand case (never was a worker, fall back), and they
  need opposite handling.

  ```sh
  if [ -f "$_dnsprefetch_self" ]; then
          # $PPID is pbr's ucode worker: the service hands the user-file wrapper
          # to "/bin/sh -c", so the shell sourcing this file is a direct child of
          # the worker, and sourcing does not fork. Confirm it rather than trust
          # it -- run by hand, $PPID is a login shell that never exits.
          _dnsprefetch_parent="$PPID"
          tr '\0' ' ' < "/proc/$PPID/cmdline" 2>/dev/null |
                  grep -q '/lib/pbr/cli\.uc' || _dnsprefetch_parent=''
          PBR_DNSPREFETCH_CHILD=1 \
          PBR_DNSPREFETCH_PARENT_PID="$_dnsprefetch_parent" \
          setsid /bin/sh "$_dnsprefetch_self" </dev/null >/dev/null 2>&1 &
  ```

  With that, an unvalidated parent yields an empty variable, rc=2 becomes
  reachable for the reason it exists, and your "worker pid unavailable, using
  ruleset fallback" message fires. Same scenario with the check in place: falls
  back, waits for the next real pbr run, resolves cleanly.

  ## 3. What the barrier is worth, measured

  To put a number on the bug you're fixing, I emulated a slow router by making
  pbr sleep between installing the ruleset and restarting dnsmasq, so its
  post-ruleset work outlasts the 15s settle — your MT6000's condition, forced.

  | | failures |
  |---|---|
  | released 1.2.3 script | **10 / 10** |
  | `b132d89` | 0 / 28 |

  The released script fails every time, and always the same way: its
  `dnsmasq_ready` is satisfied by the daemon pbr is about to restart, it fires the
  burst into the restart gap, the sets stay empty — and it still logs `[✓]`.

  ```
  pbr: Installing fw4 nft file [✓]
  pbr [12727]: Resolving domain names in policies...
  pbr [12727]: WARNING: Lookup failed for openwrt.org (timed out)   × 7
  pbr [12727]: Finished resolving 7 domain names in policies (7 failed) [✓]
  pbr: Restarting dnsmasq [✓]
  ```

  That is also the clearest argument for your `[✗]` change, which I'm now fully
  convinced by: reporting success on a total failure is the worst part of the
  current behaviour.

  I re-ran the two awkward cases here as well and they match your results: two
  overlapping reloads produce two independent runs that each follow their own
  worker and complete cleanly with no fifo left behind, and the missing-self
  branch warns and does not spawn a child. Reboot, plain reload, and
  `on_interface_reload` (correctly does not run) are all good too.

  Add the `$PPID` validation and fix that one comment, and I'll merge this.

pesa1234:
Thanks for correcting the HUP premise, and especially for validating nftset element counts and stress-testing the barrier on the DL-WRX36.

Both remaining changes are in 02e9021:

- Before detaching, the sourcing shell now reads /proc/$PPID/cmdline and only passes the PID when it matches /lib/pbr/cli.uc. A manual shell invocation therefore passes an empty PID and reaches the ruleset fallback instead of following an unrelated login shell.
- The HUP comment now states only that the second HUP is redundant on the validated worker path; the incorrect claim about an expensive adblock configuration re-read is removed.

I also checked the non-pbr-parent path directly and it selected the fallback, then reran shell syntax, diff checks, and the 71-test regression suite. All pass. A local r102 package with this final revision also builds and verifies successfully.

pesa1234:
Sorry @egc112.
One final alignment with your earlier NXDOMAIN point and the total-failure observation is in 9f0da9a:

- a partially successful run with only some NXDOMAIN results keeps [✓];
- an all-NXDOMAIN run now reports [✗], because no domain resolved into a set;
- any timeout, SERVFAIL, or other hard lookup failure still reports [✗].

The focused status matrix gives 105/106 with NXDOMAIN = success, 0/106 resolved = failure, and any hard error = failure. Shell syntax, diff checks, and the regression suite still pass.

egc112:
● Verified and merging. Thank you — three rounds of review answered in under two
  hours, each one with the change *and* the evidence for it, and you pushed back
  on nothing that deserved pushing back on. That is a pleasure to review.

  ## What I checked on hardware

  DL-WRX36, adblock ~204k domains. I emulated a slow router by making pbr sleep
  between installing the ruleset and restarting dnsmasq, so its post-ruleset work
  outlasts the old 15s settle — your MT6000's condition, forced:

  | stressed reload rounds | failures |
  |---|---|
  | released 1.2.3 script | **10 / 10** |
  | `b132d89` | 0 / 8 |
  | `9f0da9a` | 0 / 8 |

  **Parent validation (`02e9021`).** I checked the pid it follows is the real
  thing, not just a plausible number — sampling `cli.uc` processes throughout the
  run and cross-checking against what it logged:

  ```
  logged : Waiting for pbr run (pid 16902) to finish...
  alive  : 16902
  VERIFIED: pid 16902 was a live /lib/pbr/cli.uc process 
  ```

  - normal start: follows the worker, no fallback, sets go 0 -> 43 elements
  - by hand under a non-pbr shell: `worker pid unavailable, using ruleset
    fallback...`, does not follow the login shell, and completes on the next real
    pbr run — the case that sat for 46s on `b132d89`
  - two overlapping reloads: two runs, each following its own worker, both clean,
    no fifo left behind
  - boot: follows pid 5689, resolves after `Restarting dnsmasq`, 43 elements, no
    stray processes

  **Marker matrix (`9f0da9a`).** Driven straight out of `nslookup_tracker()` in
  your file rather than from the description, so this is the real code:

  ```
    all resolved                 [✓]      one hard failure        [✗]
    one NXDOMAIN among many      [✓]      all hard failures       [✗]
    several NXDOMAIN, some ok    [✓]      mixed NXDOMAIN + hard   [✗]
    all NXDOMAIN                 [✗]      single domain, NXDOMAIN [✗]
  ```

  All eight as you described, and the only behavioural difference from `b132d89`
  is the two all-NXDOMAIN rows.

  ## One trade-off, accepted deliberately

  A configuration whose *only* domain is NXDOMAIN now reports `[✗]` on every run.
  That is the steady-state cross I originally argued against, narrowed rather
  than removed — and on reflection I think you have it in the right place: if
  nothing in any policy ever reaches a set, a cross is the honest report, and the
  "one dead domain among many" case that actually bites people stays `[✓]`.
  Recording it here so it is a decision and not a surprise later.
    
  Also worth capturing for whoever reads this next: the failure this fixes is not
  merely a slow prefetch. The released script's `dnsmasq_ready` is satisfied by
  the daemon pbr is about to restart, so it fires the whole burst into the
  restart gap, leaves every set empty — and still logs `[✓]`. Silent and wrong,
  which is the worst combination.

  No `PKG_RELEASE` bump in this PR is correct; releases go out separately.

pesa1234:
Thank you very much for taking the time to review this so thoroughly and, especially, to reproduce and verify every path on real hardware. The element-count checks, stress tests, and corrections to the reasoning made the final change much stronger and gave me a lot of confidence in it.

It was genuinely a pleasure for me as well. I really appreciated the precise, constructive feedback and the care you put into validating the result. Thank you again.

PR #182, opened 2026-09-07
pbr-1.2.3: report a dns_policy whose dest_dns resolves to no server

egc112:
A `dns_policy` pointed at an interface with no determinable DNS routed nothing
and said nothing — `output.ok()`, no rules, no message.

### Why it was silent

`dest_dns` is a list of tokens sorted into three slots: the first supported
interface, the first IPv4 address, the first IPv6 address. An interface is then
*resolved*, filling whichever family slots are still empty, so downstream only
`dest_dns_ipv4` and `dest_dns_ipv6` matter. A literal address always resolves to
itself; an interface can resolve to nothing.

When it does, the filter loop pairs each source group with the DNS of its own
family and skips the ones it cannot serve:

```ucode
if (V.str_contains(fg, 'ipv4') && !dest_dns_ipv4) continue;
if (V.str_contains(fg, 'ipv6') && !dest_dns_ipv6) continue;
```

Both skips are correct — a v4 source cannot be pointed at a server that does not
exist. What was missing is any check that the loop emitted **anything**. With a
plain-address source both family groups are skipped, `dns_policy_routing()` is
never called, and the function falls through to `output.ok()`.

The error for this case already existed inside `dns_policy_routing()`, but only
`phys_dev`, `mac_address` and `domain` sources reached it — those groups carry
no family, so neither skip applies. The same misconfiguration was therefore
**reported for a MAC source and silent for an address**.

### The change

Reject it in `dns_policy_process()`, before the loop. That covers every source
kind uniformly and yields one error per policy rather than one per matching
filter group. The two `continue`s are untouched.

The check inside `dns_policy_routing()` is removed: rejecting the policy up
front makes it unreachable from both call sites, and the function is private
with no other callers. Its `dest_dns` parameter stays — the family-mismatch
error below it still names it.

### Also fixes the doubled quoting

The old call site passed ``info: "'" + dest_dns + "'"`` while the template
already quotes its argument (`"Interface '%s' has no assigned DNS"`, `pkg.uc`),
so it rendered as:

```
Interface ''wan'' has no assigned DNS
```

That was the only such site in the file. The new one passes the bare value, so
no double-quoted `info` remains anywhere.

### Scope

This is the **both-families-empty** case only. A half-resolved `dest_dns` — a
literal v6 address with a v4-only source, say — is still silently skipped. That
is a family mismatch rather than a missing server, and the existing message
would read `Interface '2606:4700:4700::1111' has no assigned DNS`, wrong on both
counts. Describing it needs a new message code, so it is left for a later
change.

No new message code and no text change here, so **no compat bump**.

### Testing

New test `04_policies/11_dns_policy_no_dns_is_reported` — 7 checks, **5/7
without this change**. It asserts the error count is exactly 2 across an
address-source and a MAC-source policy, which pins down both that the source
kind no longer decides whether the fault is reported and that the check sits
before the loop rather than inside it.

Full suite 72/72. ucode syntax gate 17/17 against the pinned interpreter.

### DIFF
--- a/files/lib/pbr/pbr.uc
+++ b/files/lib/pbr/pbr.uc
@@ -325,12 +325,6 @@ function create_pbr(fs_mod, uci_mod, ubus_mod) {
 		let nft_table = pkg.nft_table;
 		let nft_prefix = pkg.nft_prefix;
 	
-		if (!dest_dns_ipv4 && !dest_dns_ipv6) {
-			process_dns_policy_error = true;
-			push(state.errors, { code: 'errorPolicyProcessNoInterfaceDns', info: "'" + dest_dns + "'" });
-			return 1;
-		}
-	
 		// A rule built from exclusions alone has no positive source to take its
 		// family from, so fall back to the negated entries for the guards below
 		// and let the exclusions themselves decide which families to emit.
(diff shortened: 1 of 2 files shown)


egc112:
Closing — opened prematurely on my side. 1.2.3 is about to move forward with a point release, so this should be re-cut against whatever branch replaces it rather than landing here first. The work is unchanged and will come back as a fresh PR afterwards.

